PR #462, opened 2021-10-11
fix documentation

MarcAntoineSchmidtQC:
Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,7 @@
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


MarcAntoineSchmidtQC:
It did fix the issue! See [the API](https://glum--462.org.readthedocs.build/en/462/glm.html). This broke my record of most time I've spent fixing a 4 characters error.

@tbenthompson, can you confirm that this doesn't break anything? I'm not sure if the original config was important or not.

PR #492, opened 2021-11-28
Test Tweedie normalization with Wright-Bessel

### LINKED ISSUE #481: Tweedie loglikelihood with Wrights Bessel function
The normalization _constant_ of the Tweedie loglikelihood is a very complicated function, see `_tweedie_normalization` https://github.com/Quantco/glum/blob/5a2f7c7aac9dbba7de0a1c8d76aa8f046d73ae46/src/glum/_functions.pyx#L340 

Since scipy 1.7, there is the new function https://docs.scipy.org/doc/scipy-1.7.0/reference/generated/scipy.special.wright_bessel.html#scipy.special.wright_bessel which should make this computation much easier.
Is currently implemented for the range corresponding to 1<power<2.

This could potentially be a more stable approach for values of the Tweedie power close to 2 or 1.

lbittarello:
With `y > 100` or `power = 1.01`, `scipy.special.wright_bessel` returned infinities, so I couldn't test those.

Closes #481. @lorentzenchr

### DIFF
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -17,7 +17,6 @@ some parts and tricks stolen from other sklearn files.
 from __future__ import division
 
 import copy
-import logging
 import sys
 import warnings
 from collections.abc import Iterable
@@ -60,8 +59,6 @@ from ._solvers import (
     _trust_constr_solver,
 )
 from ._util import _align_df_categories, _safe_toarray
-
-_logger = logging.getLogger(__name__)
 
 _float_itemsize_to_dtype = {8: np.float64, 4: np.float32, 2: np.float16}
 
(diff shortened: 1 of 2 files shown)


MarcAntoineSchmidtQC (approved):
lgtm

lorentzenchr:
Just as info, I did not open an issue in scipy because of the following:
```python
from scipy.special import wright_bessel

y = 6
power = 1.999
dispersion = 1
alpha = (2 - power) / (1 - power)  # -0.0010010010010008906
x = (((power - 1) / y) ** alpha) / ((2 - power) * (dispersion ** (1 - alpha)))  # 1001.7961657008032

wright_bessel(-alpha, 0, 0.7 * x)
```
gives
```
2.7524526105751842e+304
```
And note that
```python
import numpy as np
np.finfo(np.float64).max
```
gives
```
1.7976931348623157e+308
```
Conclusion: no obvious flaw in `wright_bessel`

PR #502, opened 2021-12-28
Add daily unit tests

BenBarrettQC:
Closes #429.

I've set this up for Linux so far (pending your feedback I will add macOS and Windows) and issues are opened on CI failure (see e.g. https://github.com/Quantco/glum/issues/499).

Some additional comments:
- Some tests against the nightly builds of pandas, numpy and scikit-learn are currently running into errors due to [this](https://github.com/scikit-learn/scikit-learn/pull/21880), which gets imported upstream [here](https://github.com/dask/dask-ml/blob/30b3deab19755ecd45156ef9cac9fee646615100/dask_ml/model_selection/_search.py#L26).
- Other tests against the nightly builds are failing [here](https://github.com/Quantco/glum/blob/007d82675912d6605724a48af7d1ec80e4329141/src/glum/_glm.py#L1085) because the option context is finding duplicate keys; I suspect we just need to add e.g. `"display."` to `max_rows` etc.
- The other daily tests are currently passing (i.e. non-nightly tests), see e.g. [here](https://github.com/Quantco/glum/commit/f085a23f268856a7cedab5bdef536bc5580be20e).
- Once you're happy with the PR I'll comment in the code section that opens issues on CI failure; I wanted to avoid spamming everyone for now.
- #498 to follow.

Checklist
* [x] Added a `CHANGELOG.rst` entry

PR #509, opened 2022-01-03
Consolidate actions

jtilly:
Quick follow-up PR to #502. This merges `daily_unittests/actions.yml` into `unittests/actions.yml` and `test_daily.sh` into `test.sh`.


BenBarrettQC (approved):
Nice work.

PR #513, opened 2022-01-19
changed exact pins for minimum versions

MarcAntoineSchmidtQC:
I don't know exactly why we pinned sphinx to a specific version. @tbenthompson maybe you remember?

I think this might solve the problem with the new readthedocs PR.

Checklist

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -43,10 +43,10 @@ dependencies:
   - make
   - matplotlib-base
   - nbclassic>=0.2.8
-  - nbsphinx=0.8.3
+  - nbsphinx>=0.8.3
   - openml # used for downloading datasets in the tutorials.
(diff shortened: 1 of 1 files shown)


MarcAntoineSchmidtQC:
see #512

tbenthompson:
I think there was a temporary problem with that version of Sphinx that has likely been long since solved. But I don't remember precisely!

jtilly:
Other projects have similar issues, see, e.g. https://github.com/microsoft/LightGBM/pull/4957#issuecomment-1017124668.

astrojuanlu:
Alternatively, pinning `python=3.9` in `environment.yml` might do the trick? Right now Python 3.10 is being picked up.

PR #519, opened 2022-02-14
New feature: add information criteria for model diagnostics

### LINKED ISSUE #516: Model diagnostics - AIC, BIC
Please can we consider adding some model diagnostics for comparing two or more models. I was thinking AIC and BIC.  I think the ln likelihood is available so it should be straightforward to add these as attributes Statsmodel has these attributes e.g. model_results.AIC()

NicholasHoernleQC:
**Checklist:**
* [x] Decision on what to do with ridge/elastic net regularisation. See [here](https://stats.stackexchange.com/questions/564563/how-to-calculate-the-degrees-of-freedom-for-l1-and-l2-regularised-glms) for more details.
* [x] Decision on  what to do with CV implementation. **Decision:** information criteria are primarily a metric for estimating out of sample generalisation from in sample measures. If cross validation is applicable to the setting (i.e., there is enough data) it does not seem to make sense to use these metrics so we have not supplied them on the GeneralizedLinearRegressorCV class. 

Closes #516.

**Summary:**
This change implements the calculations of the `aic`, `aicc` and `bic` information criteria for the trained model. I placed these criteria as properties/attributes on the `glum.GeneralizedLinearRegressor` class.

**Notes:**
- These information criteria require an "effective number of parameters" of the model. In the case of an unregularized model, this is simply the count of features/parameters. In the case of a lasso model, this is the [count of the non-zero parameters](https://projecteuclid.org/journals/annals-of-statistics/volume-35/issue-5/On-the-degrees-of-freedom-of-the-lasso/10.1214/009053607000000127.full) on the ML fit. In the case of a ridge/elastic net model, I could not find a consensus on how this should be implemented so I opted to add a warning here. Does anyone have a better suggestion or is there something that I have missed? 
 **Decision:** I have answered this [in more detail here](https://stats.stackexchange.com/a/564869/245707) but I will argue that it does not make sense to compute these values for models that use L2 regularisation. I think keeping the warning there when L2 regularisation is used is sufficient to highlight this issue.
- The method to compute these scores is always called from the `fit` method of `glum.GeneralizedLinearRegressor`. This is because we require the `X` and `y` data sources to compute these values. This allows for a simple interface where the score is defined as an attribute of the trained regressor, e.g: `regressor.aic`, `regressor.aicc` or `regressor.bic`. This could rather be done as a separate method call that accepts `X` and `y` as arguments: `regressor.aic(X, y)`, `regressor.aicc(X, y)` or `regressor.bic(X, y)`. I opted for the first choice as (1) I think this is neater and (2) it makes sense to compute these values at train time; but I am happy to change to the alternative. 
  **Decision:** We have decided not to compute these statistics at train time due to the unnecessary computational overhead on the fit method. Rather, passing the dataset at call-time seems preferable.
- These information criteria are only available when the noise model is one of BinomialDistribution, GammaDistribution, NormalDistribution, PoissonDistribution. See line `1640`. This is because we require the definition of the likelihood function. Maybe I missed them but are there other families that we should include here? 
  **Decision:** include the TweedieDistribution that generalises the above. 
- Calling `aic`, `aicc` or `bic` on the model *before* it has been trained returns None. We could rather log a warning or throw an exception. I am not sure if there is a standard/preferred approach here?  
  **Decision:** raise error.


NicholasHoernleQC:
I think merging main back into this PR once [513](https://github.com/Quantco/glum/pull/513) is in will fix the docs error.

CHANGELOG.rst
```diff
@@ -7,6 +7,15 @@
 Changelog
 =========
 
+2.0.5 - 2022-02-14
```
lbittarello, line 10:
```suggestion
2.0.5 - 2022-XX-XX
```

src/glum/_glm.py
```diff
@@ -1552,6 +1625,53 @@ def score(
...
+        # we require that the log_likelihood be definied
+        if not isinstance(
+            self.family_instance,
+            (
+                BinomialDistribution,
+                GammaDistribution,
+                NormalDistribution,
+                PoissonDistribution,
+            ),
+        ):
```
lbittarello, line 1648:
```suggestion
        if not isinstance(
            self.family_instance, (BinomialDistribution, TweedieDistribution)
        ):
```

src/glum/_glm.py
```diff
@@ -1552,6 +1625,53 @@ def score(
...
+            return False
+
+        ddof = np.sum(np.abs(self.coef_) > np.finfo(self.coef_.dtype).eps)
+
+        X = self._info_criteria["X"]
+        y = self._info_criteria["y"]
```
lbittarello, line 1654: Why not just take `X` and `y` as arguments?

lbittarello (commented):
> The method to compute these scores is always called from the `fit` method

What if I want to compute out-of-sample values?

src/glum/_glm.py
NicholasHoernleQC: Hey Luca. So these information criteria are typically defined as in sample measures only. So if you do actually have out of sample data (e.g., validation or test) then it is much better to use standard cross-validation or just optimisation on the validation set using held out log-likelihood. My suggestion is to force this best practice by not giving `X` and `y` as arguments -- i.e., if someone finds they are computing these on the validation set something has gone wrong.

src/glum/_glm.py
NicholasHoernleQC: See answer above. These are *in sample* metrics that are used to approximate out of sample error (they do this by penalising the complexity of the model). So I wanted to make this method run when the `fit` method was called, thereby forcing that in-sample data be used.

src/glum/_glm.py
NicholasHoernleQC: I can make the change above but there are two things to note:

1. The `log_likelihood` throws a `NotImplementedError` for the `InverseGaussianDistribution` (TweedieDistribution(p=3)). So we'd need a try catch here to catch that exception.
2. The `log_likelihood` method appears to hang for values of p \in (1,2). I think this has to do with the series approximation that is mentioned in the docs. This might be a bug that I can create an issue for but it will cause this method to hang indefinitely for now. This can be managed by checking for self.power in [0,1,2] but this is going to be a more verbose piece of code. What do you think?

MarcAntoineSchmidtQC:
> I think merging main back into this PR once [513](https://github.com/Quantco/glum/pull/513) is in will fix the docs error.

Yes, I will merge this right now. With such a small PR we can skip the PR review.

src/glum/_glm.py
lbittarello: > My suggestion is to force this best practice by not giving `X` and `y` as arguments

The function is private anyway, so I don't know that the benefits outweigh the considerably more complicated implementation.

> we'd need a try catch here to catch that exception.

You can also check the `power` attribute of Tweedie distributions and skip the information criteria if it isn't in the [0, 2] interval.

> The `log_likelihood` method appears to hang for values of p in (1,2). 

Strange. We have tests ([here](https://github.com/Quantco/glum/blob/b072bd652a8c24295acadde63c9253f262bc3d71/tests/glm/test_distribution.py#L400) and [here](https://github.com/Quantco/glum/blob/b072bd652a8c24295acadde63c9253f262bc3d71/tests/glm/test_distribution.py#L468)), for a variety of parametrizations, and we don't run into deadlocks. Perhaps you're running into parallelization issues? Can you run those tests locally? Seems worth investigating.

NicholasHoernleQC: > The function is private anyway, so I don't know that the benefits outweigh the considerably more complicated implementation.

Yep thats a good point. I'll update this. 

> You can also check the power attribute of Tweedie distributions and skip the information criteria if it isn't in the [0, 2] interval.

I'll make this change as well.

> Strange. We have tests ([here](https://github.com/Quantco/glum/blob/b072bd652a8c24295acadde63c9253f262bc3d71/tests/glm/test_distribution.py#L400) and [here](https://github.com/Quantco/glum/blob/b072bd652a8c24295acadde63c9253f262bc3d71/tests/glm/test_distribution.py#L468)), for a variety of parametrizations, and we don't run into deadlocks. Perhaps you're running into parallelization issues? Can you run those tests locally? Seems worth investigating.

The tests work as is. However, if I change the `method` argument of the dispersion calculation to default ( method="pearson") then the tests hangs ([line 406](https://github.com/Quantco/glum/blob/b072bd652a8c24295acadde63c9253f262bc3d71/tests/glm/test_distribution.py#L406)). I think it has to do with the dispersion being really small in the tests (`8.213159509438382e-23`). It might be some kind of underflow or something. I am still not too sure why it hangs though.

MarcAntoineSchmidtQC:
> These information criteria require an "effective number of parameters" of the model. In the case of an unregularized model, this is simply the count of features/parameters. In the case of a lasso model, this is the [count of the non-zero parameters](https://projecteuclid.org/journals/annals-of-statistics/volume-35/issue-5/On-the-degrees-of-freedom-of-the-lasso/10.1214/009053607000000127.full) on the ML fit. In the case of a ridge/elastic net model, I could not find a consensus on how this should be implemented so I opted to add a warning here. Does anyone have a better suggestion or is there something that I have missed? See line 764.

For ridge, [this source](https://arxiv.org/pdf/1509.09169.pdf) and [this source](https://www.researchgate.net/publication/238242580_Ridge_regression_and_its_degrees_of_freedom) seem to agree that if we ignore the parameter tuning of the penalty, the degree of freedom for ridge is very similar to OLS with a small correction for the regularization.

For elastic net, it seems logical that we can combine the two approaches (number of non-zero coefficient with a correction for the l2 regularization). However, I can't find any source backing this conjecture.

NicholasHoernleQC:
> Yes, I will merge this right now. With such a small PR we can skip the PR review.

Brilliant. Thanks. I have merged main back into this branch. Do you prefer it if these feature branches are rebased or is it okay to merge main in like I have done?

MarcAntoineSchmidtQC:
Merging main into the branch is perfectly fine.

src/glum/_glm.py
```diff
@@ -718,6 +718,59 @@ def link_instance(self) -> Link:
...
+    def get_info_criteria(self, crit: str):
+        if hasattr(self, "_info_criteria"):
+            if (self.alpha is not None and self.alpha > 0) or self.l1_ratio < 1.0:
+                warnings.warn(
+                    "There is no robust definition for the model's degrees of "
+                    + f"freedom under L2 (ridge) regularisation. The {crit} "
+                    + "might not be well defined in these cases."
+                )
+            return self._info_criteria[crit]
+        return None
```
MarcAntoineSchmidtQC, line 772: I think it would make more sense to raise an error if the model as not been fitted before (so that we don't have the `_info_criteria` attribute). This would be similar to trying to predict using a model that has not been fitted.

src/glum/_glm.py
```diff
@@ -718,6 +718,59 @@ def link_instance(self) -> Link:
...
             return get_link(self.link, self.family_instance)
 
+    @property
+    def aic(self):
+        """
+        Akaike's information criteria. Computed as:
+        :math:`-2\\log\\hat{\\mathcal{L}} + 2\\hat{k}` where
+        :math:`\\hat{\\mathcal{L}}` is the maximum likelihood estimate of the
+        model, and :math:`\\hat{k}` is the effective number of parameters under
+        an assumption of lasso regularisation. In practice, this is the
```
MarcAntoineSchmidtQC, line 728: Great job with the docstrings!

src/glum/_glm.py
```diff
@@ -1552,6 +1616,53 @@ def score(
...
+            self.family_instance, TweedieDistribution
+        ) and self.family_instance.power not in [0, 1, 2]:
+            return False
+
+        ddof = np.sum(np.abs(self.coef_) > np.finfo(self.coef_.dtype).eps)
+
+        k_params = ddof + self.fit_intercept
+
+        mu = self.predict(X)
+        (nobs,) = mu.shape
```
MarcAntoineSchmidtQC, line 1651:
```suggestion
        nobs = mu.shape[0]
```
This will work no matter how many dimensions mu has.
It might work right now but I would like to support multi-class models in the near future.

src/glum/_glm.py
```diff
@@ -831,6 +884,17 @@ def _tear_down_from_fit(self):
         del self._solver
         del self._random_state
 
+    def _set_up_for_info_criteria(self, **kwargs):
```
MarcAntoineSchmidtQC, line 887: I'm not sure I understand the need for this method. It's only called once in the `fit` method and will always have the same argument (`y`).

MarcAntoineSchmidtQC:
> I placed these criteria as properties/attributes on the glum.GeneralizedLinearRegressorBase class but it might be preferable to have them on the glum.GeneralizedLinearRegressor class. Let me know if you'd like this to be changed?

I think it's at the right location in `GeneralizedLinearRegressorBase` but it is currently not implemented for `GeneralizedLinearRegressorCV`. You would need to make the same changes you have made to `GeneralizedLinearRegressor.fit` inside of `GeneralizedLinearRegressorCV.fit`.

With `GeneralizedLinearRegressorCV`, I think it would be preferable to also have a warning saying that these measure do not take into account the fact that the cross-validation is data-driven.

MarcAntoineSchmidtQC (commented):
Thanks a lot for the PR! Some preliminary comments. Let's discuss the big picture implementation of this before I give another review.

NicholasHoernleQC:
> With `GeneralizedLinearRegressorCV`, I think it would be preferable to also have a warning saying that these measure do not take into account the fact that the cross-validation is data-driven.

Yes perfect. I was wondering about this one and forgot to mention it in the summary. On the one hand it makes sense to have there but on the other hand the whole point of CV is to avoid using these in sample estimators so not too sure these scores belong.

src/glum/_glm.py
lbittarello: As a curiosity: this would also work: `nobs, *_ = mu.shape`.

lbittarello: A more interesting point: If the weights represent frequency weights, `nobs` should be the sum of weights instead of the number of rows. If they represent probability weights, the notion of degrees of freedom becomes more complicated.

src/glum/_glm.py
```diff
@@ -2283,8 +2393,8 @@ def fit(
         -------
         self
         """
-
         self._validate_hyperparameters()
+        X_in, sample_weights_in = X, sample_weight
```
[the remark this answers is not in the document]
lbittarello: It isn't very elegant, but I don't see a way around it. Maybe @MarcAntoineSchmidtQC has an idea? As it is, I don't think you need to worry about the weights (they're only normalized), unless you want to use them to determine the effective number of observations.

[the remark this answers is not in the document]
NicholasHoernleQC: It might come down to the discussion on whether or not this *should* be run as a default at fit time or should rather be executed in a lazy manner. This links to the benchmarking that @MarcAntoineSchmidtQC mentioned below but if we choose to only run this `_compute_information_criteria` when someone calls `aic`/`bic` etc then we'd either need to revert to saving `X` and `y` on the object (tearing down again after computation) or we take X and y as arguments to aic/bic etc (which I would not vote for).

src/glum/_glm.py
```diff
@@ -2474,3 +2478,172 @@ def fit(
...
+
+        y : array-like, shape (n_samples,)
+            Same data as used in 'fit'
+
+        sample_weight : array-like, shape (n_samples,), optional (default=None)
+             Same data as used in 'fit'
+        """
+        return self.get_info_criteria("bic", X, y, sample_weight)
+
+    def get_info_criteria(
```
MarcAntoineSchmidtQC, line 2627: This could be a private method (start with an underscore) so that we don't pollute the namespace. If it makes sense to keep it public, then I would add a docstring.

src/glum/_glm.py
```diff
@@ -2474,3 +2478,172 @@ def fit(
...
+
+        mu = self.predict(X)
+        ll = self.family_instance.log_likelihood(y, mu, sample_weight=sample_weight)
+
+        aic = -2 * ll + 2 * k_params
+        bic = -2 * ll + np.log(nobs) * k_params
+        if nobs > k_params + 1:
+            aicc = aic + 2 * k_params * (k_params + 1) / (nobs - k_params - 1)
+        else:
+            aicc = None
```
MarcAntoineSchmidtQC, line 2546: If aicc can be None, I would raise an error when someone tries to call it instead of returning None.

tests/glm/test_glm.py
```diff
@@ -1788,3 +1789,49 @@ def test_score_method(as_data_frame, offset, weighted):
...
     # use pytest because NumPy used to always reject comparisons against zero
     assert pytest.approx(score, 1e-8) == int(offset is not None)
+
+
+def test_information_criteria(regression_data):
+    X, y = regression_data
+    regressor = GeneralizedLinearRegressor(family="normal", alpha=0)
+    regressor.fit(X, y)
+    assert np.allclose(
+        [138.80, 141.59, 168.20],
```
MarcAntoineSchmidtQC, line 1799: Are these values taken from a previous run?
If possible, I would test the results against statsmodels.

MarcAntoineSchmidtQC (changes requested):

tests/glm/test_glm.py
NicholasHoernleQC: They are taken from a statsmodels run on the same data. I can include a note there that they are from statsmodels or just include the statsmodels fit in the test?

MarcAntoineSchmidtQC (approved):
Looks good! Thanks a lot Nicholas.

peterlee18:
Great job guys. Being using GLUM extensively (unpenalised for now via inputting the design matrix) and its working great. It fits extremely quickly on AWS as it leverages all the cores available.

PR #520, opened 2022-02-15
Reimplementation of PR #512 with latest main branch

MarcAntoineSchmidtQC:
See PR #512 for background. This PR only purpose is to update PR #512 with the latest changes to main.

Checklist
* [ ] ~~Added a `CHANGELOG.rst` entry~~


MarcAntoineSchmidtQC:
Approving this because it's only a re-write of #512

PR #522, opened 2022-02-19
Update jinja

### LINKED ISSUE #521: Daily run failure: Unit tests
The daily unit tests failed. See https://github.com/Quantco/glum/actions/runs/1867838974 for details.

MarcAntoineSchmidtQC:
This should fix #521 

Checklist

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -37,7 +37,7 @@ dependencies:
 
   # documentation dev
   - altair # used in docs/tutorials/rossman
-  - jinja2<3
+  - jinja2
   - jupyterlab
   - jupytext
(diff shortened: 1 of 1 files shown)


MarcAntoineSchmidtQC:
Markupsafe, a dependency of jinja, removed a function in a new version. See https://github.com/pallets/markupsafe/pull/261/files and https://github.com/pallets/jinja/issues/1585. This problem has been fixed in newer versions of jinja but we were pinning jinja to < 3. I don't think the reason to pin jinja is still an issue so I'm removing the pin.

jtilly (approved):

PR #523, opened 2022-02-19
Badges and docs

MarcAntoineSchmidtQC:
I added badges and I clarified that glum was developed for problems with N >> K in the docs.

Checklist
* [ ] ~~Added a `CHANGELOG.rst` entry~~


jtilly (approved):
Looks fancy!

I don't think we should advertise that we're supporting python 3.6 (I guess the information is pulled from https://github.com/Quantco/glum/blob/7fec22462cd171f13e2ce3e03b7f591830c745a1/setup.py#L62).

Also, it's about time to start testing against 3.10 and add that to the package metadata. But that's for a different PR.

PR #526, opened 2022-02-23
Allow calculation of standard errors with categorical data

### LINKED ISSUE #514: std_errors method
I fitted a model using the non-cv class and tried to extract std errors using the provided method. I tried both Poisson on claim freq and Tweedie on cost per vehicle and neither was successful.

My x_factors contain a mix of categorical and continuous factors which don't need to be transformed to numerics or one-hot when calling the fit and predict methods. However, the std_errors method does not accept the untransformed x_factors. Should I expect the method to work on the original x_factors with cat factors as strings etc.

I also tried reducing my model to one continuous factor only - this ran successfully but the std errors coming out didn't look very plausible. Is there something amiss with the std_errors method. The parameters (coeffs) themselves all look plausible.

By the way, this is a fantastic package and fits my 1.7m record dataset in less than 10 seconds whilst statsmodel takes 45 minutes. So thank you team for putting this together. I'm pretty sure that this will be the go to package for GLMs, particularly as you iron out the inevitable issues for a new package.

DavidMeesterQC:
Closes #514

This PR allows the `std_errors` and `covariance_matrix` method of `glum.GeneralizedLinearRegressor` to accept a `pandas.DataFrame`, now allowing the same `X` data used in fitting the model.

In particular, the calculation of the covariance matrix did not allow for all `tabmat` matrices. When X is a  `tabmat.CategoricalMatrix`, `tabmat.SplitMatrix`, or `pandas.DataFrame` with categorical or sparse data, the matrices used for further calculations were not supported in the calculation of the standard errors. 

Further, this PR fixes a small errror where the nonrobust standard errors of a GLM with the binomial family and logit link were calculated incorrectly. The input argument of `get_one_over_variance` only affected [a calculation](https://github.com/Quantco/glum/blob/5a910716388f743f91c4828a322681708b5af302/src/glum/_distribution.py#L1169) for the binomial distribution with logit link.

Tests have been added for the calculation of standard errors with categorical data and dataframes. Tests for the standard errors of additional models, with the binomial distribution (with logit link) and poisson (with log link) have also been added.

**Note:**
I have kept the current test of the standard errors against statsmodels' OLS for now, but since I am also testing a GLM with Gaussian distribution and identity link, we may consider removing that one. I built most of the tests based on the `regression_data` fixture, but I am happy to restructure them as well!

Checklist
* [x] Add further tests for different types of input data (dataframes, `tabmat` matrices)

### DIFF
--- a/src/glum/_distribution.py
+++ b/src/glum/_distribution.py
@@ -5,7 +5,7 @@ from typing import Tuple, Union
 import numexpr
 import numpy as np
 from scipy import sparse, special
-from tabmat import MatrixBase, StandardizedMatrix
+from tabmat import CategoricalMatrix, MatrixBase, SplitMatrix, StandardizedMatrix
 
 from ._functions import (
     binomial_logit_eta_mu_deviance,
@@ -457,7 +457,7 @@ class ExponentialDispersionModel(metaclass=ABCMeta):
             Whether the model has an intercept.
         """
         W = (link.inverse_derivative(link.link(mu)) ** 2) * get_one_over_variance(
-            self, link, mu, link.inverse(mu), dispersion, sample_weight
+            self, link, mu, link.link(mu), dispersion, sample_weight
         )
 
         return _safe_sandwich_dot(X, W, intercept=fit_intercept)
@@ -522,12 +522,16 @@ class ExponentialDispersionModel(metaclass=ABCMeta):
         ).reshape(-1, 1)
 
         if fit_intercept:
-            if sparse.sputils.isdense(X):
+            if not (
+                sparse.issparse(X) or isinstance(X, (SplitMatrix, CategoricalMatrix))
+            ):
                 return np.hstack((W, np.multiply(X, W)))
             else:
                 return sparse.hstack((W, X.multiply(W)))
         else:
-            if sparse.sputils.isdense(X):
+            if not (
+                sparse.issparse(X) or isinstance(X, (SplitMatrix, CategoricalMatrix))
+            ):
                 return np.multiply(X, W)
             else:
                 return X.multiply(W)
--- a/src/glum/_distribution.py
+++ b/src/glum/_distribution.py
@@ -5,7 +5,13 @@ from typing import Tuple, Union
 import numexpr
 import numpy as np
 from scipy import sparse, special
-from tabmat import MatrixBase, StandardizedMatrix
+from tabmat import (
+    DenseMatrix,
+    MatrixBase,
+    SparseMatrix,
+    SplitMatrix,
+    StandardizedMatrix,
+)
 
 from ._functions import (
     binomial_logit_eta_mu_deviance,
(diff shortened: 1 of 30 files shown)


src/glum/_distribution.py
```diff
@@ -522,15 +528,35 @@ def _score_matrix(self, link, X, y, mu, sample_weight, dispersion, fit_intercept
...
-            if sparse.issparse(X):
-                return sparse.hstack((W, X.multiply(W)))
-            else:
+            if sparse.sputils.isdense(X):
                 return np.hstack((W, np.multiply(X, W)))
-        else:
-            if sparse.issparse(X):
-                return X.multiply(W)
             else:
+                if isinstance(X, SplitMatrix):
```
MarcAntoineSchmidtQC, line 534: This is the right idea, but I think it should be implemented in tabmat and not here. What we need here is a simple element-wise multiplication. Some classes already implement the `multiply` method. Let's implement them for the other classes so that here we can simply call `X.multiply(W)`

src/glum/_glm.py
```diff
@@ -1472,13 +1477,19 @@ def covariance_matrix(
             if clusters is not None:
                 n_groups = len(np.unique(clusters))
                 grouped_gradient = _group_sum(clusters, gradient)
-                inner_part = grouped_gradient.T @ grouped_gradient
+                if not isinstance(grouped_gradient, tm.SplitMatrix):
+                    inner_part = grouped_gradient.T @ grouped_gradient
+                else:
+                    inner_part = grouped_gradient.sandwich(np.ones_like(y))
```
MarcAntoineSchmidtQC, line 1483: I think it would be best if we avoided making a separate computation for a SplitMatrix and for other types. Actually, I don't think `grouped_gradient` can be a SplitMatrix. We should check this.

MarcAntoineSchmidtQC (commented):
Let's move a lot of the logic from this PR to tabmat. Ideally, glum would never make a distinction between sparse, dense, categorical, and split-matrices. I'll work on a PR to tabmat tomorrow.

src/glum/_glm.py
DavidMeesterQC: Yes, the gradient could be a SplitMatrix and after `_group_sum` this is an array, I have updated this here. For other calculations we will have to update `tabmat` to remove seperate computations for a SplitMatrix.

MarcAntoineSchmidtQC:
@DavidMeesterQC, I just released a new version of tabmat with `multiply` capabilities and a `drop_first` argument for `CategoricalMatrix`. Let me know if you need help integrating those in this PR.

FYI, it might take a few hours for the new tabmat version to become available on conda-forge.

lbittarello:
@MarcAntoineSchmidtQC, @DavidMeesterQC, what's up here?

DavidMeesterQC:
> @MarcAntoineSchmidtQC, @DavidMeesterQC, what's up here?

This is ready for review 😄

src/glum/_glm.py
```diff
@@ -629,9 +629,16 @@ def is_pos_semidef(p: Union[sparse.spmatrix, np.ndarray]) -> Union[bool, np.bool
...
-    out = np.empty((len(np.unique(groups)), data.shape[1]))
-    for i in range(data.shape[1]):
-        out[:, i] = np.bincount(groups, weights=data[:, i])
+    ngroups = len(np.unique(groups))
+    out = np.empty((ngroups, data.shape[1]))
+    if sparse.issparse(data) or isinstance(
+        data, (tm.SplitMatrix, tm.CategoricalMatrix)
+    ):
+        for i in range(data.shape[1]):
+            out[:, i] = (np.eye(ngroups)[:, groups] @ data.getcol(i)).ravel()
```
lbittarello, line 638: Could we create `np.eye(ngroups)` before starting the `for` loop so we only need to create it once?

lbittarello: I'm taking it that `np.bincount(groups, weights=data.getcol(i))` doesn't work?

tests/glm/test_glm.py
```diff
@@ -1735,6 +1747,53 @@ def test_std_errors(regression_data):
...
+@pytest.mark.parametrize("fit_intercept", [True, False])
+@pytest.mark.parametrize("family", ["poisson", "normal", "binomial"])
+def test_array_std_errors(regression_data, classification_data, family, fit_intercept):
+    X, y = regression_data
+    if family == "poisson":
+        y = np.round(abs(y))
+        sm_family = sm.families.Poisson()
+        dispersion = 1
+    elif family == "binomial":
+        X, y = classification_data
```
lbittarello, line 1759: Can't you just transform the outcome to be binary here (as for Poisson) instead of adding a fixture?

lbittarello (approved):
Thanks, @DavidMeesterQC!

lbittarello (approved):

PR #537, opened 2022-04-28
Fix issue 536

MarcAntoineSchmidtQC:
This fixes issue #536. The issue was caused by an update in git and while it's been fixed in recent version of checkout, there are no solution when using a container to run the next commands.

Checklist


jtilly (approved):
Windows CI failure is unrelated. 

xref https://github.com/rpy2/rpy2/issues/874

PR #542, opened 2022-05-13
Add `fisher_diag` option, as well as option to make state.hessian a sparse COO matrix

AlanTuQC:
Allows the `libsvmdata` finance dataset to run without memory issues. However, the sandwich product is still very slow, so we are looking into other methods (e.g. less precise Hessian calculation, first-order instead of second-order).

### DIFF
--- a/src/glum/_cd_fast.pyx
+++ b/src/glum/_cd_fast.pyx
@@ -209,7 +209,8 @@ def enet_coordinate_descent_gram(int[::1] active_set,
 def enet_coordinate_descent_gram_diag_fisher(
                                  #floating[:,:] X, 
                                  #floating[:] hessian_rows,
-                                 np.ndarray[np.float64_t, ndim=2] X,
+                                 #np.ndarray[np.float64_t, ndim=2] X,
+                                 X,
                                  np.ndarray[np.float64_t, ndim=1] hessian_rows,
                                  int[::1] active_set,
                                  floating[::1] w,
@@ -258,10 +259,12 @@ def enet_coordinate_descent_gram_diag_fisher(
     cdef np.ndarray[np.float64_t, ndim=1] Q_active_set_ii = np.empty(n_active_features, dtype=np.float64)
     # is empty bad?
 
-    # used to check correctness of code - compare with Q from non-diag-fisher
-    cdef np.ndarray[np.float64_t, ndim=2] Q_check = np.empty((n_active_features, n_active_features), dtype=np.float64)
+    cdef cur_col  # makes sure we can overwrite this and not take up too much memory?
 
-    # THE inefficiency here is that we must re-calculate the rows of Q for ever iteration.
+    # used to check correctness of code - compare with Q from non-diag-fisher
+    # cdef np.ndarray[np.float64_t, ndim=2] Q_check = np.empty((n_active_features, n_active_features), dtype=np.float64)
+
+    # THE inefficiency here is that we must re-calculate the rows of Q for every iteration.
 
     # with nogil:
     for n_iter in range(max_iter):
@@ -286,10 +289,20 @@ def enet_coordinate_descent_gram_diag_fisher(
             if active_set[0] < <unsigned int>intercept:
                 if ii == 0:
                     Q_active_set_ii[0] = hessian_rows.sum()
-                    Q_active_set_ii[1:] = np.matmul(hessian_rows, X)
+                    Q_active_set_ii[1:] = X.transpose_matvec(hessian_rows)
+                    # Q_active_set_ii[1:] = np.matmul(hessian_rows, X)  
                 else:
-                    Q_active_set_ii[0] = np.dot(hessian_rows, X[:, active_set_ii - 1])
-                    Q_active_set_ii[1:] = np.matmul((hessian_rows * X[:, active_set_ii - 1]), X)
+                    cur_col = X[:, active_set_ii - 1]
+                    Q_active_set_ii[0] = cur_col.transpose_matvec(hessian_rows)
+                    Q_active_set_ii[1:] = X.transpose_matvec(cur_col.multiply(hessian_rows).ravel())
+
+                    # Q_active_set_ii[0] = X[:, active_set_ii - 1].transpose_matvec(hessian_rows)
+                    # Q_active_set_ii[1:] = X.transpose_matvec(X[:, active_set_ii - 1].multiply(hessian_rows).ravel())
+
+                    # Q_active_set_ii[1:] = X.transpose_matvec(hessian_rows * X[:, active_set_ii - 1].A.ravel())
+                    # Q_active_set_ii[1:] = X.transpose_matvec(hessian_rows * X.A[:, active_set_ii - 1]) 
+                    # Q_active_set_ii[0] = np.dot(hessian_rows, X[:, active_set_ii - 1])
+                    # Q_active_set_ii[1:] = np.matmul((hessian_rows * X[:, active_set_ii - 1]), X)  # use matvec
 
             # if ii < <unsigned int>intercept:
             #     # Q_active_set_ii[1:] = np.matmul(hessian_rows, X[:, 1:])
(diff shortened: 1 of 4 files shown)


AlanTuQC:
`fisher_diag=True` means that the entire Hessian is never stored at one time; instead, we calculate the Hessian rows "on demand," using each row before overwriting it. This leads to more operations being done, but it saves on memory.

src/glum/_solvers.py
```diff
@@ -151,9 +182,25 @@ def update_hessian(state, data, active_set):
             np.arange(data.X.shape[0], dtype=np.int32),
             active_set,
         )
-        state.hessian[np.ix_(active_set, active_set)] = hessian_init
+
+        # In the sparse Hessian case, state.hessian is a coo_matrix.
+        # hessian_init is a numpy array of size active_set x active_set;
+        # we can convert this to a coo_matrix and simply add it to the state.hessian
+        if state.use_sparse_hessian:
+            coo_data = hessian_init.flatten()
```
BenBarrettQC, line 190: Does it make sense to explicitly pass the flattening order here? I worry that if default behavior under the hood were to change then your repeat / tiling operations would yield invalid indices.

src/glum/_solvers.py
```diff
@@ -176,13 +224,32 @@ def update_hessian(state, data, active_set):
...
 
-    return (
-        state.hessian[np.ix_(active_set, active_set)],
-        n_active_rows,
-    )
+    # If state.hessian is in COO form, we have to convert to CSC in order to index the
+    # active set elements, which we then convert to a numpy array for compatibility with
+    # the rest of the code
+    if state.use_sparse_hessian:
+        # this is a little awkward
```
BenBarrettQC, line 246:
```suggestion
```

src/glum/_solvers.py
```diff
@@ -176,13 +224,32 @@ def update_hessian(state, data, active_set):
...
+        return (
+            state.hessian.tocsc()[np.ix_(active_set, active_set)].toarray(),
+            n_active_rows,
+        )
```
BenBarrettQC, line 250: I may be missing something, but are we at all at risk here of returning to the domain where we run out of memory because we cast the sparse Hessian to a dense matrix?

src/glum/_solvers.py
```diff
@@ -265,13 +332,14 @@ def _irls_solver(inner_solver, coef, data) -> Tuple[np.ndarray, int, int, List[L
 
     Repeat steps 1-3 until convergence.
     Note: Use hessian matrix instead of Hessian for H.
+        This used to say "Use Fisher matrix instead of Hessian for H."
```
BenBarrettQC, line 335: Do we need this addendum / history?

src/glum/_solvers.py
```diff
@@ -528,9 +603,19 @@ def __init__(self, coef, data):
         self.old_hessian_rows = np.zeros(data.X.shape[0], dtype=data.X.dtype)
         self.gradient_rows = None
         self.hessian_rows = None
-        self.hessian = np.zeros(
-            (self.coef.shape[0], self.coef.shape[0]), dtype=data.X.dtype
-        )
+
+        # In order to avoid memory issues for very wide datasets (e.g. Issue #485), we
+        # can instantiate the Hessian as a sparse COO matrix.
+        self.use_sparse_hessian = True
```
BenBarrettQC, line 609: This shouldn't be hardcoded / there should be some flag to allow users to change the behavior.

src/glum/_cd_fast.pyx
```diff
@@ -204,6 +205,139 @@ def enet_coordinate_descent_gram(int[::1] active_set,
...
+    cdef floating max_min_subgrad
+    cdef unsigned int active_set_ii, active_set_jj
+    cdef unsigned int ii, jj
+    cdef int n_iter = 0
+    cdef unsigned int f_iter
+    cdef UINT32_t rand_r_state_seed = rng.randint(0, RAND_R_MAX)
+    cdef UINT32_t* rand_r_state = &rand_r_state_seed
+
+    # Equivalent to Q[active_set_ii] in the diag_fisher=False version
+    cdef np.ndarray[np.float64_t, ndim=1] Q_active_set_ii = np.empty(n_active_features, dtype=np.float64)
```
MarcAntoineSchmidtQC, line 253: Can you use memoryviews instead of the old numpy interface? The rest of the code uses memoryviews so this would also make it more consistent.

src/glum/_cd_fast.pyx
```diff
@@ -204,6 +205,139 @@ def enet_coordinate_descent_gram(int[::1] active_set,
...
+
+    cdef cur_col  # Does the cdef here (rather than using it as a local variable) reduce memory?
+    
+    # nogil is incompatible with our slicing, numpy, and sparse matrix operations, but 
+    # may be necessary to increase performance
```
MarcAntoineSchmidtQC, line 258: I think memoryviews will solve this issue.

src/glum/_glm.py
```diff
@@ -2168,6 +2183,7 @@ def __init__(
         A_ineq: Optional[np.ndarray] = None,
         b_ineq: Optional[np.ndarray] = None,
         force_all_finite: bool = True,
+        diag_fisher=False,
```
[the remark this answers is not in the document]
MarcAntoineSchmidtQC: yes, once we settle on a syntax we should also apply this to `_glm_cv.py`.

src/glum/_cd_fast.pyx
```diff
@@ -140,7 +141,7 @@ def enet_coordinate_descent_gram(int[::1] active_set,
         for n_iter in range(max_iter):
             w_max = 0.0
             d_w_max = 0.0
-            for f_iter in range(n_active_features):  # Loop over coordinates
```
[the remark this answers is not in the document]
AlanTuQC: No -- I got confused about which comments I had made (i.e. notes to self) and which ones were already there. Will put it back

src/glum/_cd_fast.pyx
```diff
@@ -111,14 +111,15 @@ def enet_coordinate_descent_gram(int[::1] active_set,
...
         Q = X^T X (Gram matrix)
         q = X^T y
     """
 
-    # get the data information into easy vars
     cdef unsigned int n_active_features = active_set.shape[0]
-    cdef unsigned int n_features = Q.shape[0]
+    cdef unsigned int n_features = Q.shape[0]  # are you sure this is correct?
+    # n_features is exactly now the same as n_active_features; should it be q.shape[0]?
+    # however, the variable is never used again, so I guess we are okay
```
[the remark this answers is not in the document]
AlanTuQC: I was hesitant to delete because the variable is defined in the original code and perhaps the authors intended to use it for something later

src/glum/_cd_fast.pyx
```diff
@@ -204,6 +205,139 @@ def enet_coordinate_descent_gram(int[::1] active_set,
...
+    cdef unsigned int ii, jj
+    cdef int n_iter = 0
+    cdef unsigned int f_iter
+    cdef UINT32_t rand_r_state_seed = rng.randint(0, RAND_R_MAX)
+    cdef UINT32_t* rand_r_state = &rand_r_state_seed
+
+    # Equivalent to Q[active_set_ii] in the diag_fisher=False version
+    cdef np.ndarray[np.float64_t, ndim=1] Q_active_set_ii = np.empty(n_active_features, dtype=np.float64)
+
+    cdef cur_col  # Does the cdef here (rather than using it as a local variable) reduce memory?
```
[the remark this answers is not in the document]
AlanTuQC: Yes, but `cur_col` is of type SparseMatrix (will double check this), and at least in the original Cython code, we do not specify the types of sparse matrices

src/glum/_glm.py
```diff
@@ -672,6 +672,7 @@ def __init__(
...
```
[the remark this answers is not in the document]
AlanTuQC: Yes -- I saw that the inclusion of type hints is inconsistent throughout the arguments, so was not sure which to prefer

src/glum/_glm.py
```diff
@@ -969,13 +971,15 @@ def _solve(
...
                 verbose=self.verbose > 0,
+                diag_fisher=self.diag_fisher,
             )
             if self._solver == "irls-ls":
                 coef, self.n_iter_, self._n_cycles, self.diagnostics_ = _irls_solver(
                     _least_squares_solver, coef, irls_data
                 )
             # 4.2 coordinate descent ##############################################
             elif self._solver == "irls-cd":
+                # [Alan] This is the case we're concerned with for wide problems.
```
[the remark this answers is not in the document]
AlanTuQC: Not sure what best practice is for this -- I don't think it's a necessary comment for those familiar with the code, but may still be helpful to orient me

src/glum/_solvers.py
```diff
@@ -151,9 +182,25 @@ def update_hessian(state, data, active_set):
             np.arange(data.X.shape[0], dtype=np.int32),
             active_set,
         )
-        state.hessian[np.ix_(active_set, active_set)] = hessian_init
+
+        # In the sparse Hessian case, state.hessian is a coo_matrix.
+        # hessian_init is a numpy array of size active_set x active_set;
```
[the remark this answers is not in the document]
AlanTuQC: correct

src/glum/_cd_fast.pyx
AlanTuQC: On hold

src/glum/_cd_fast.pyx
AlanTuQC: On hold

AlanTuQC:
This PR has been split into two: one focused on the `use_sparse_hessian` feature, and another focused on the `diag_fisher` feature.

Issue #485, opened 2021-11-19
MemoryError when fitting on sparse X as apparently a Hessian matrix is being instantied ? 


mathurinm:
requirement: `pip install libsvmdata`  -- python utility to download data from LIBSVM website; the first time downloading the data may take 2 mins.

The following script causes a MemoryError on my machine:
```python
from libsvmdata import fetch_libsvm
from glum import GeneralizedLinearRegressor

X, y = fetch_libsvm("finance")
clf = GeneralizedLinearRegressor(family='gaussian', l1_ratio=1, alpha=1).fit(
    X, y)
```

output:
```python
---------------------------------------------------------------------------
MemoryError                               Traceback (most recent call last)
~/workspace/mem_error.py in <module>
      3 
      4 X, y = fetch_libsvm("finance")
----> 5 clf = GeneralizedLinearRegressor(family='gaussian', l1_ratio=1, alpha=1).fit(
      6     X, y)

~/miniconda3/lib/python3.8/site-packages/glum/_glm.py in fit(self, X, y, sample_weight, offset, weights_sum)
   2448                     )
   2449                 )
-> 2450             coef = self._solve(
   2451                 X=X,
   2452                 y=y,

~/miniconda3/lib/python3.8/site-packages/glum/_glm.py in _solve(self, X, y, sample_weight, P2, P1, coef, offset, lower_bounds, upper_bounds, A_ineq, b_ineq)
    976             # 4.2 coordinate descent ##############################################
    977             elif self._solver == "irls-cd":
--> 978                 coef, self.n_iter_, self._n_cycles, self.diagnostics_ = _irls_solver(
    979                     _cd_solver, coef, irls_data
    980                 )

~/miniconda3/lib/python3.8/site-packages/glum/_solvers.py in _irls_solver(inner_solver, coef, data)
    287     https://www.csie.ntu.edu.tw/~cjlin/papers/l1_glmnet/long-glmnet.pdf
    288     """
--> 289     state = IRLSState(coef, data)
    290 
    291     state.eta, state.mu, state.obj_val, coef_P2 = _update_predictions(

~/miniconda3/lib/python3.8/site-packages/glum/_solvers.py in __init__(self, coef, data)
    529         self.gradient_rows = None
    530         self.hessian_rows = None
--> 531         self.hessian = np.zeros(
    532             (self.coef.shape[0], self.coef.shape[0]), dtype=data.X.dtype
    533         )

[...]

PR #543, opened 2022-05-19
Add `use_sparse_hessian` functionality to prevent a dense Hessian from being instantiated

AlanTuQC:
When `use_sparse_hessian=True`, `state.hessian` is stored as a sparse COO matrix. Partially addresses Issue #485 -- the dataset with 4M+ columns can now run without a memory error.

### DIFF
--- a/src/glum/_cd_fast.pyx
+++ b/src/glum/_cd_fast.pyx
@@ -209,7 +209,8 @@ def enet_coordinate_descent_gram(int[::1] active_set,
 def enet_coordinate_descent_gram_diag_fisher(
                                  #floating[:,:] X, 
                                  #floating[:] hessian_rows,
-                                 np.ndarray[np.float64_t, ndim=2] X,
+                                 #np.ndarray[np.float64_t, ndim=2] X,
+                                 X,
                                  np.ndarray[np.float64_t, ndim=1] hessian_rows,
                                  int[::1] active_set,
                                  floating[::1] w,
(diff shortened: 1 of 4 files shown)


abudis:
Hey Alan, thanks for looking into this! Is this planned to be merged in the near future? I'm experiencing a similar problem with a large sparse matrix.

lbittarello:
Please reopen when it's ready for review.

zmbc:
My data is not much wider than it is long -- it is roughly square (and large). My sense is that this PR, if completed and merged, would make glum work for me -- or should I expect that glum's solvers wouldn't work even if they could be run?

Even if not, @tbenthompson said on https://github.com/Quantco/glum/issues/485 that

> the implementation of a solver more appropriate for [K >> N] would not be a huge undertaking and 75% of the pieces necessary already exist within glum

What is missing? I'd be happy to contribute something here, as I have yet to find a Python library that can handle this case without regularization.

zmbc:
I started on this in #869, but I ran into this: https://github.com/Quantco/tabmat/issues/17. I wonder if @ElizabethSantorellaQC might be able to offer some guidance. The text in that issue seems to mostly be focused on how to _tell_ whether or not a sandwich product will be sparse, but even the option for the user to say "I know it will be, represent it as a sparse matrix" would be enough for my purposes.

stanmart:
Thank you so much for looking into this feature! I agree, as a first step I would not worry about using heuristics to try to guess whether the hessian will be sparse, but just rely on the user telling glum whether it should be. (Btw I doubt Elizabeth will chime in on this as she hasn't worked on glum in a while.)

PR #544, opened 2022-05-19
Add `diag_fisher` functionality to allow the Hessian to be computed "on the fly"

AlanTuQC:
Instead of storing the entire Hessian matrix, `diag_fisher=True` calculates Hessian rows _only when they must be used_ (i.e. in the relevant step of coordinate descent). It accomplishes this by storing the current state of the 1-D array `hessian_rows` (which appears in the middle of the sandwich product). This necessitates a greater number of matrix operations (many of them redundant), but also saves memory.

To-Do: 
- `state.hessian` does not need to be instantiated if `diag_fisher=True`; this is not yet accounted for in the current code.
- Make sure we've addressed all entry points where the argument `diag_fisher` can be set by the user.
- How do we implement memory views (and possibly `nogil`) in `_cd_fast.pyx`?

### DIFF
--- a/diag_fisher_benchmarks.py
+++ b/diag_fisher_benchmarks.py
@@ -1,0 +1,35 @@
+# %%
+
+# %load_ext autoreload
+# %autoreload 2
+
+# %%
+
+import numpy as np
+from libsvmdata import fetch_libsvm
+
+from glum import GeneralizedLinearRegressor
+
+X, y = fetch_libsvm("finance")
+
+# %%
+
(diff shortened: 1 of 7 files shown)


MarcAntoineSchmidtQC (changes requested):
Really good job. Before merging I think we should run some benchmarks and post them in the PR thread. It would also be good to add to the documentation.

lbittarello:
Please reopen when it's ready for review.

Issue #496, opened 2021-12-20
GLM fit without penalty


lorentzenchr:
I would be very nice to also fit completely unpenalized GLMs (as in base R glm). For categorical features, one then needs to drop one level, called reference or base level.
```python
from plotnine.data import diamonds
from glum import GeneralizedLinearRegressor


# Use the 4 c's as feature
# Note that cut, color, clarity have dtype category
X = diamonds.loc[:,["carat", "cut", "color", "clarity"]]

# Targets
y = diamonds["price"]

glm = GeneralizedLinearRegressor(alpha=0, family="gamma", link="log")
glm.fit(X, y)
```
gives
```
LinAlgError: Matrix is singular.
```
The error is correct as no categorical level was dropped.

jtilly:
xref https://github.com/Quantco/tabmat/issues/75

jtilly:
In applications where we need to drop a base level, we typically one-hot encode our categoricals before using glum. With few levels, that's also faster than using categorical types. As you can see in the [tabmat issue](https://github.com/Quantco/tabmat/issues/75), we have discussed this before but we were a bit unsure what a good API design for this. The easiest solution would be to drop the first category per variable. Do you think that would be good enough?

lorentzenchr:
AFAIK, R also drops the first level so this seems like a good default, for sure, only in case of `alpha=0`.
If someone still needs more control, she/he could go for `sklearn.preprocessing.OneHotEncoder(drop=...)` or for a formula solution.

peterlee18:
Now tabmat has support to drop the first column of a CategoricalMatrix, are there any plans to leverage this to allow direct fitting of unpenalised GLMs (i.e. without creating the design matrix prior to input)?

lorentzenchr:
That‘s great. Thank you for all your work!

PR #571, opened 2022-09-21
Use "drop_first" keyword to drop first category in categorical column

NicholasHoernleQC:
**Summary**
Uses [implementation in tabmat](https://github.com/Quantco/tabmat/pull/168) to create a new argument for `GeneralizedLinearRegressorBase` to drop the first category in a categorical column.

This permits to run unregularised models.

~~There were a number of tests that needed adjusting, this is because the number of categories in the tests changed. We mostly set the `drop_first=False` argument in these cases. For the `tests/glm/test_glm.py::test_categorical_types` we, rather passed the `drop_first=True` to `pd.get_dummies` to create a baseline.~~

In the case that a categorical column contains only one category, and `drop_first=True` is set, we raise a `ValueError`. 

Joint author with @MatthiasLoefflerQC :)

**Updates:**

- We added a change to `_expand_categorical_penalties` to ensure that the P1 and P2 arguments are compatible with drop_first. See: d308605

**Closes** #496 

**Checklist**


src/glum/_glm.py
```diff
@@ -208,6 +211,20 @@ def _check_offset(
     return offset
 
 
+def _name_categorical_variables(
+    categories: Tuple[str], column_name: str, drop_first: bool
+):
+    new_names = [
+        f"{column_name}__{category}" for category in categories[1 * drop_first :]
```
MarcAntoineSchmidtQC, line 218:
```suggestion
        f"{column_name}__{category}" for category in categories[int(drop_first):]
```

explicit

src/glum/_glm.py
```diff
@@ -208,6 +211,20 @@ def _check_offset(
...
+    ]
+    if len(new_names) == 0:
+        raise ValueError(
+            f"Categorical column: {column_name}, contains only one category. "
+            + "This should dropped from the feature matrix."
```
MarcAntoineSchmidtQC, line 223:
```suggestion
            + "This should be dropped from the feature matrix."
```

src/glum/_glm.py
```diff
@@ -1274,6 +1294,7 @@ def predict(
             copy=self._should_copy_X(),
             ensure_2d=True,
             allow_nd=False,
+            drop_first=True,
```
MarcAntoineSchmidtQC, line 1297:
```suggestion
            drop_first=self.drop_first,
```

Can you otherwise explain why drop_first should be enforced here?

src/glum/_glm.py
```diff
@@ -2181,6 +2210,7 @@ def __init__(
         A_ineq: Optional[np.ndarray] = None,
         b_ineq: Optional[np.ndarray] = None,
         force_all_finite: bool = True,
+        drop_first: bool = True,
```
MarcAntoineSchmidtQC, line 2213: Same comment as above.

src/glum/_glm_cv.py
```diff
@@ -303,6 +307,7 @@ def __init__(
         force_all_finite: bool = True,
         cv=None,
         n_jobs: Optional[int] = None,
+        drop_first: bool = True,
```
MarcAntoineSchmidtQC, line 310: Same comment as above.

MarcAntoineSchmidtQC (changes requested):
Great work! See my comments below.

MarcAntoineSchmidtQC:
>  and improves lasso performance with categorical variables.

Have you documented the performance improvement with lasso? I would be curious to see the numbers.

src/glum/_glm.py
```diff
@@ -680,6 +697,7 @@ def __init__(
...
```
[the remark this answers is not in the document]
MatthiasLoefflerQC: Hi Marc,

that is a good point and worthwhile discussing as there are pro's and con's to both
possibilities.

When using a penalized estimator and ``drop_first=False`` all categories in a categorical column are penalized equally. This is intuitive and makes sense to us. By contrast, if the first category is dropped, the first category is only implicitly defined through the other categories (and the intercept) and hence not penalized at the same level as the other categories.

On the other hand, when fitting the lasso, both the theoretical performance, as well as the numerical optimization performance, depend on the condition number of the feature matrix. When fitting an intercept (which is True by default) and having drop_first=False, the feature matrix is singular and the condition number unbounded. This, from a theory perspective, suggests to use ``drop_first = True`` as default when fitting the lasso.

However, Nick and I ran a lot of experiments this morning to try to validate this. For example, we ran tests with categorials in the feature matrix that were both correlated and uncorrelated with the dependent variable. Unexpectedly the results were very similar for both cases of drop_first = True/False. Importantly, the optimization did not appear to be unstable.

Hence, we agree with you to keep ``drop_first = False`` as default, as the only usecase of ``drop_first = True `` seems to be ``alpha=0``and ``solver = 'auto' `` and changed things accordingly.

NicholasHoernleQC:
> Have you documented the performance improvement with lasso

We removed this claim. After some experimentation this morning, we couldn't turn the theory into results.

MarcAntoineSchmidtQC (approved):
Sorry for the delayed review. This looks very good! Thanks a lot for the contribution.

PR #578, opened 2022-10-19
Bounds with constant features

MarcAntoineSchmidtQC:
Fixes an issue when a column of `X` would be constant. Setting bounds and inequalities would not work because they are being scaled by the standard deviation of the column.

I've also removed two internal variables (`_solver` and `_center_predictors`) from being deleted at the end of the fitting routine. They are useful for advanced users to understand which solver was used when the input is 'auto'.

Checklist

### DIFF
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -327,17 +327,17 @@ def _standardize(
     X, col_means, col_stds = X.standardize(sample_weight, center_predictors, True)
 
     if col_stds is not None:
+        penalty_mult = _one_over_var_inf_to_val(col_stds, 1.0)
         # We copy the bounds when multiplying here so the we avoid
         # side effects.
         if lower_bounds is not None:
-            lower_bounds = lower_bounds * col_stds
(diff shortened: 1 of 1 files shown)


lbittarello (approved):
Wonderful, thanks! Could you also fix the [typo on L2104](https://github.com/Quantco/glum/blob/5509e25f6bb1b90610093deb702fd58e7c2973e0/src/glum/_glm.py#L2104)?

src/glum/_glm.py
```diff
@@ -327,17 +327,17 @@ def _standardize(
     X, col_means, col_stds = X.standardize(sample_weight, center_predictors, True)
 
     if col_stds is not None:
+        penalty_mult = _one_over_var_inf_to_val(col_stds, 1.0)
```
PhilippRuchser, line 330: Now that we use `penalty_mult` in various places, maybe a different name would be appropriate (such as `col_stds_inv`) - at least I was confused by the term "penalty".

PhilippRuchser (approved):
LGTM, Thanks!

PR #580, opened 2022-10-20
fixing pandas 1.6.0 and 2.0.0 breaking changes

### LINKED ISSUE #575: Daily run failure: Unit tests
The daily unit tests failed. See https://github.com/Quantco/glum/actions/runs/3232211996 for details.

MarcAntoineSchmidtQC:
In the development version of pandas, `get_dummies` generates boolean data, and when used in arithmetic operations, this becomes a numpy array with dtype python object.

Glum was actually not the problem. When using a pandas dataframe as inputs, tabmat takes care to cast everything to the right dtype. The problem was with the tests themselves where they would pandas dataframes with `.values` to acces the underlying numpy array.

Coincidentally, 2 days ago pandas released the 2.0.0 development version. Since we started seeing the problem before that, it seems to be happening with 1.6.0 and 2.0.0.

This PR fixes #575.

**TODO: remove daily CI on push**

Checklist


jtilly:
The `get_dummies` dtype change was introduced in 1.6.0 and classified as bugfix, see https://github.com/pandas-dev/pandas/pull/48022#pullrequestreview-1111560850.

tests/glm/test_glm.py
```diff
@@ -1827,7 +1827,7 @@ def test_dataframe_std_errors(regression_data, categorical, split, fit_intercept
         alpha=0, family="normal", fit_intercept=fit_intercept
     )
     mdl.fit(X=X, y=y)
-    X_sm = pd.get_dummies(X)
+    X_sm = pd.get_dummies(X).to_numpy(dtype=float)
     if fit_intercept:
```
jtilly, line 1831: `get_dummies` also accepts a dtype argument.

jtilly (approved):

bashtage:
I agree that this change is problematic.  It caused the canary build of statsmodels to also start failing.  I think this change unfortunately is going to cause a lot of problems, and it should seriously be considered for rolling back.  The `np.asarray(df.get_dummies())` returning object is a very hard to debug problem.  Even `df.get_dummies().to_numpy()` produces a not-useful dtype.

bashtage:
Oops.  I should have made most of my case on the pandas tracker.

PR #586, opened 2022-11-26
Ci update nov 2022

MarcAntoineSchmidtQC:
Checklist


MarcAntoineSchmidtQC:
No really sure why conda-build was failing. However, since conda-forge is also using "mambabuild", I like this approach, especially since we are trying to predict a conda-forge failure here.

Regarding the miniforge version update, is this something dependabot can do in the future, @xhochy?

xhochy:
> Regarding the miniforge version update, is this something dependabot can do in the future, @xhochy?

Not with Dependabot but we could setup a custom Github Actions workflow that runs on a weekly basis that checks for updates.

PR #590, opened 2022-12-09
Use tm.from_pandas for any pd.dataframe

### LINKED ISSUE #589: Use SplitMatrix if dataframe contains sparse columns
Currently, `_set_up_and_check_fit_args` returns a DenseMatrix if the entering dataframe contains dense and sparse columns, yet no categoricals anymore after they have been dummy encoded ([runs into else here](https://github.com/Quantco/glum/blob/main/src/glum/_glm.py#L1737-L1741), end ultimately ends up [here](https://github.com/Quantco/glum/blob/main/src/glum/_glm.py#L1863-L1864)). Instead, we could analogously check if the dataframe contains any sparse columns and then use `X=tm.from_pandas(X)` (we wouldn't need `drop_first` as dataframe already transformed). 

Happy to set up a PR after having agreed on the best way to implement this. @lbittarello, feel free to chip in with further suggestions.

ChristianRoerigQC:
Closes #589. This PR suggests that we call `tabmat.from_pandas` on any dataframe. In the case that it does not contain any categoricals, we don't drop any levels. 
 
Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,11 @@ Changelog
 UNRELEASED
 ----------
 
+**Bug fixes:**
+
+- A dataframe with dense and sparse columns was transformed to a dense matrix instead of a split matrix by :meth:`~glum.GeneralizedLinearRegressor._set_up_and_check_fit_args`.
+  Fixed by calling ``tabmat.from_pandas`` on any dataframe.
+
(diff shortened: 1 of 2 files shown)


MarcAntoineSchmidtQC (approved):
Thanks! Perfect simple implementation.

PR #616, opened 2023-03-14
Don't use march="native" on arm64 

### LINKED ISSUE #607: `-march` is not cross-plaform
Same issue as in https://github.com/Quantco/tabmat/issues/223

The setup script tries to set argument `-march=native`:
https://github.com/Quantco/glum/blob/a405fd73bdb959f533358027201ebb641a60f700/setup.py#L37

But this argument is not accepted by GCC in platforms like ARM and PPC. Instead, those platforms use `mcpu` (at least for GCC in linux), so to achieve the same thing, it should use `-mcpu=native`.

jtilly:
Closes #607.

Not trying to set `march='native'` anymore on arm64. Same as in [tabmat]( https://github.com/Quantco/tabmat/blob/11531bcae57086226d8a4508a92b406819fef8cf/setup.py#L105-L110).

Also: We're now running Windows and macOS CI when `setup.py` or one of the environment YAMLs get changed.

(Note: CI for macOS is currently failing)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ channels:
 dependencies:
   # required for users (note: this is not where you specify new dependencies
   # for the conda packages. please put those `conda.recipe/meta.yaml`!!
-  - libblas>=0=*mkl  # uncomment for m1
+  - libblas>=0=*mkl  # comment for m1
   - numexpr
   - pandas
   - tabmat>=3.1.0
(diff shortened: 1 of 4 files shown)


environment.yml
```diff
@@ -5,7 +5,7 @@ channels:
 dependencies:
   # required for users (note: this is not where you specify new dependencies
   # for the conda packages. please put those `conda.recipe/meta.yaml`!!
-  - libblas>=0=*mkl
+  - libblas>=0=*mkl  # uncomment for m1
```
xhochy, line 8:
```suggestion
  - libblas>=0=*mkl  # comment for m1
```

xhochy (approved):

jtilly:
Let's merge #617 instead.

PR #618, opened 2023-03-15
Build wheels for Python 3.11

jtilly:
We have [tabmat wheels](https://pypi.org/project/tabmat/#files) for 3.11.


xhochy:
This fails to build as we are building  `scipy` from source in win32. I would propose to skip `*-win32`.

```
              Collecting scipy>=1.3.2
                Downloading scipy-1.10.1.tar.gz (42.4 MB)
```

xhochy (approved):

PR #623, opened 2023-03-27
Cython lint pre commit

adityagoel4512:
This adds a pre-commit hook to lint cython files, checking for unused imports, unused variables etc.


lbittarello (approved):
🚀

xhochy:
Should we set this up as a conda-based hook so that we can also use it in restricted environment?

adityagoel4512:
> Should we set this up as a conda-based hook so that we can also use it in restricted environments?

That seems like a good idea - do you have any pointers on how to do this?

xhochy (approved):

jtilly (approved):

PR #641, opened 2023-05-24
Remove the dask_ml dev dependency

MartinStancsicsQC:
Checklist

## Description

`dask_ml` was only used to create the benchmark datasets. The few required classes and functions from `dask_ml.preprocessing` (`Categorizer`, `DummyEncoder`, `OrdinalEncoder`) and `dask_ml.compose` (`ColumnTransformer`, `make_column_transformer`) has been included added to the `glum` codebase with appropriate modifications.

## Testing

 - Tests in `test_benchmark_golden_master.py` are passing 🎉
 - The benhmark suite still works for `glum` (`glm_benchmarks_run --library_name glum`)
 - Some of the benchmarks fail for `h2o` (e.g. `narrow-insurance-no-weights-l2-binomial`). However, it is not a result of this PR as the same happens in the latest released version (aa5ad98). Maybe there was a change in the `h2o` api (its version is not pinned)? *Should I create a separate issue for this?*

## Questions

 - should we have unit tests for these new functions? ~~(there are no tests for the benchmark as of yet, apart from tests of the cli)~~
 - Is `src/glum_benchmarks/data/categorical_util.py` the right place to put these classes?


jtilly:
Thanks for this!

- Could we get rid of the entire `categorical_util` module? I would think that the calls to the `Categorizer` can be replaced with `pd.get_dummies` and that most of the other classes in that module could probably be replaced by their scikit-learn equivalents.
- Note that we can't really get rid of the dask-ml dependency, because it's still used in the docs, e.g., [here](https://github.com/Quantco/glum/blob/c9eb22d8149f96f40cbe1cefc2ab8a80882f6c0e/docs/tutorials/glm_french_motor_tutorial/glm_french_motor.ipynb#L49).

MarcAntoineSchmidtQC:
If I recall correctly, the only reason we used dask_ml was that it retained the pandas dataframe output structure (as opposed to sklearn which transformed your data into a numpy array). Since we wrote that part of the code, sklearn has been supporting dataframe output, so maybe we can only rely on sklearn now. See [original PR](https://github.com/Quantco/glum/pull/231).

I agree with Jan that adding the `categorical_util` module is not optimal. 

> * Note that we can't really get rid of the dask-ml dependency, because it's still used in the docs, e.g., [here](https://github.com/Quantco/glum/blob/c9eb22d8149f96f40cbe1cefc2ab8a80882f6c0e/docs/tutorials/glm_french_motor_tutorial/glm_french_motor.ipynb#L49).

It would be great to have a separate environment file for this purpose and use it [here](https://github.com/Quantco/glum/blob/c9eb22d8149f96f40cbe1cefc2ab8a80882f6c0e/.readthedocs.yml#L13).

MartinStancsicsQC:
I'm not a fan of having a separate module either, it feels overkill for just transforming a couple of benchmark datasets. However, I think `sklearn` outputs `pd.DataFrame`s only since version 1.2 (October 2022), so we would need to pin a rather recent version. Also, AFAIK it still has no support for categoricals, so `OrdinalEncoder` and `Categorizer` would need to be implemented to some extent anyway.

Regarding the docs, I really like the idea of having a separate environment file. It's already done for the benchmarks so it would not be unprecedented either.

MartinStancsicsQC:
> One way to do this is to have a separate environment file for the docs and use it [here](https://github.com/Quantco/glum/blob/c9eb22d8149f96f40cbe1cefc2ab8a80882f6c0e/.readthedocs.yml#L13).

It seems to me that rerunning the example notebooks is currently not part of the build process, so pointing to a different environment file here does not really influence notebook execution. Because updating the examples is done manually, one solution would be to create this new `environment-docs.yml` file, and explain in the contributing section that people wishing to update/change the examples need do `mamba update --name glum --file environment-docs.yml` (just like how it's done for the benchmarks).

A slightly unrelated question/idea: if we incorporated rerunning those notebooks into the docs build process, and then we could at least monitor if the examples are still up-to-date or if they break for some reason. However this would mean that `dask-ml` would have to remain a dev dependency as the package could not be built without it. Also a failure to run the examples would lead to a build failure, which I'm not sure if that's desirable or not.

MarcAntoineSchmidtQC:
> It seems to me that rerunning the example notebooks is currently not part of the build process

Right, that make sense. 

Personally, I think that I would rather remove the dask-ml dependency than adding the notebooks in the docs build process. If we simply want to make sure it is still running, we could either create a new CI step (independent of the docs) that check whether we run into any issues with the rendering of the notebooks.

MartinStancsicsQC:
I got rid of the separate module. The `DummyEncoder` class could be replaced by simply using `pd.get_dummies`, and the rest were only used in the `src/glum_benchmarks/data/create_insurance.py` module, so I moved them there. Also deleted a lot of unnecessary checks/functionality, so I think they are of a much more reasonable length now, and are just thin wrappers over pandas functions.

The only ugly part that remains is handling old pandas versions which had worse support for categorical data. If I can pin `pandas >= 0.21` (ca. 2017) in the dev dependencies, then that part goes away, too. I don't think that any `glum` user has an older `pandas` than that.

MarcAntoineSchmidtQC:
Go ahead and add a pandas version constraint. 0.21 is totally fine.

MarcAntoineSchmidtQC (approved):
Looks good. Thanks!

PR #650, opened 2023-06-01
Feature decollinearizer

MatthiasSchmidtblaicherQC:
FYI: branch is not ready for review yet, turning this into a draft PR in order to allow commenting on code.

Checklist

### DIFF
--- a/src/glum/_transformers.py
+++ b/src/glum/_transformers.py
@@ -166,4 +166,6 @@ class Decollinearizer(TransformerMixin, BaseEstimator):
                     == self.column_mapping[col_idx].category,  # noqa: B023
                     col_name,
                 ] = self.column_mapping[col_idx].base_category
+        for col in X.select_dtypes(include="category").columns:
+            X[col] = X[col].cat.remove_unused_categories()
         return X.drop(columns=cols_to_drop)
--- a/src/glum/_transformers.py
+++ b/src/glum/_transformers.py
@@ -27,7 +27,8 @@ def _find_collinear_columns_pandas(
     use_tabmat: bool = True,
     tolerance=1e-6,
 ) -> CollinearityResults:
-    """Decompose a pandas.DataFrame into a tabmat.SplitMatrix.
+    """Find the collinear columns and categories in a pandas DataFrame.
+    Return all information needed to decollinearize the data.
 
     Parameters
     ----------
@@ -82,7 +83,8 @@ def _find_collinear_columns_pandas(
 def _find_collinear_columns_numpy(
     X: np.ndarray, fit_intercept=True, mode="gram", tolerance: float = 1e-6
 ) -> CollinearityResults:
-    """Decompose a numpy.ndarray into a tabmat.SplitMatrix.
+    """Find the collinear columns in a numpy array.
+    Return all information needed to decollinearize the array.
 
     Parameters
     ----------
--- a/src/glum/_transformers.py
+++ b/src/glum/_transformers.py
@@ -130,6 +130,11 @@ def _find_collinear_columns(
         The indices of the columns to keep and the columns to drop.
     """
     permuted_keep_mask = np.abs(np.diag(R)) > tolerance
+    # More columns than rows case:
+    if R.shape[1] > R.shape[0]:
+        permuted_keep_mask = np.concatenate(
+            (permuted_keep_mask, np.zeros(R.shape[1] - R.shape[0], dtype=bool))
+        )
     keep_mask = np.empty_like(permuted_keep_mask)
     keep_mask[P] = permuted_keep_mask
 
--- a/src/glum/_transformers.py
+++ b/src/glum/_transformers.py
@@ -1,6 +1,6 @@
 """Some utilities for transforming data before fitting a model."""
 
-from typing import Hashable, List, NamedTuple, Optional, Sequence, Tuple, Union
+from typing import Any, Hashable, List, NamedTuple, Optional, Sequence, Tuple, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


src/glum/_transformers.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def _find_collinear_columns_pandas(
+    df: pd.DataFrame,
+    fit_intercept=True,
+    mode: str = "gram",
+    use_tabmat: bool = True,
+    tolerance=1e-6,
+) -> CollinearityResults:
+    """Decompose a pandas.DataFrame into a tabmat.SplitMatrix.
```
MatthiasSchmidtblaicherQC, line 30: Does not seem to be the main purpose of this function.

src/glum/_transformers.py
```diff
@@ -0,0 +1,347 @@
...
+        raise ValueError(f"Mode must be 'gram' or 'direct', got {mode}")
+        # Cannot happen
+
+    return _find_collinear_columns(R, P, fit_intercept, tolerance=tolerance)
+
+
+def _find_collinear_columns_numpy(
+    X: np.ndarray, fit_intercept=True, mode="gram", tolerance: float = 1e-6
+) -> CollinearityResults:
+    """Decompose a numpy.ndarray into a tabmat.SplitMatrix.
```
MatthiasSchmidtblaicherQC, line 85: See above.

src/glum/_transformers.py
```diff
@@ -0,0 +1,347 @@
...
+            a tabmat.SplitMatrix, or a numpy.ndarray.
+        y: Optional[VectorLike]
+            Ignored. Present for API consistency.
+
+        Returns
+        -------
+        Self
+            The fitted transformer.
+        """
+        if mode not in ["gram", "direct", None]:
```
MatthiasSchmidtblaicherQC, line 222: It would be useful to add a test that the product of modes QR and gram and numpy and tabmat ways all give the same result. I would also be happy with reducing the number of options in case there is a favorite. E.g, if gram works well everywhere and is faster for `k << n`, then we can scrap the direct mode.

src/glum/_transformers.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+class Decollinearizer(TransformerMixin, BaseEstimator):
+    """Drop collinear columns from the design matrix implied by a dataset.
+
+    The type of the output is the same as the input. For non-categorical
+    columns, collinear ones are simply dropped. For categorical columns
+    (e.g. in a pandas.DataFrame or a tabmat.SplitMatrix), values whose
+    columns in the design matrix would be dropped are replaced with the
+    first category. This supposes that the first category is the reference,
```
MatthiasSchmidtblaicherQC, line 186: no Oxford comma with two items :)

MatthiasSchmidtblaicherQC (commented):
Thanks, looks like good progress! 🚀 

High level, some next steps would be:
1. Check if a mode appears superior to the other (likely gram for `k << n`) and focus on perfecting this one. For this, benchmarks on artificial or real data would be very helpful.
2. Think about where this functionality best fits. If we keep it as transformer, it might better fit in a downstream package. But we might also want it as part of the `GeneralizedLinearRegressor`. To decide on the latter, we can zoom a bit out and think about how unregularized GLMs fit in here.
3. Compare the dropping behaviour to what some other GLM packages do.

src/glum/_transformers.py
MartinStancsicsQC: True, fixed

src/glum/_transformers.py
MartinStancsicsQC: Agreed. I'll create some tests that check the modes against each other. I'll also do some benchmarks and then we can decide how to proceed.

MartinStancsicsQC:
> 1. Check if a mode appears superior to the other (likely gram for `k << n`) and focus on perfecting this one. For this, benchmarks on artificial or real data would be very helpful.

Agreed, I'll create some benchmarks. My hunch is that there might be a time-memory trade-off here. For data that is already one-hot encoded, the direct method might be faster, while for datasets with many categorical columns, the gram-matrix-based method avoids blowing up the size of the data. I agree that the numpy-gram combination probably does not make much sense and probably should be removed (let's wait for benchmarks, though).

MartinStancsicsQC:
Current state as of [b25fc72](https://github.com/Quantco/glum/pull/650/commits/b25fc72e6410d410e37ebc7d696be46faacd9a32):

 - Everything is based on the `X'X` matrix, even in the dense input case. It turns out that doing it is *both* faster and more memory-efficient than calculating the QR decomposition of `X` directly.
 - The intercept is never dropped. This is consistent with the behavior of other statistical packages I've used.

Still a lot to do:
 - Major decisions
   - Should it be a transformer?
   - Should it be in `glum`?
   - Should we have a convenience function that combines decollinearization (probably the first use of this word ever) and estimation?
   - See if the choice of dropped columns has some pattern, and if it is similar to other packaegs.
 - Additional checks and stuff
   - Check things about `pd.DataFrame` before converting to `tabmat` matrix (like in `_glm.py`)
   - More checks in general? (finite values, non-zero rows/columns, etc.)
   - Fix some type signatures (e.g. `ArrayLike` is too general)
   - is `_safe_sandwich_dot` a private function in `_utils`?
 - Testing
   - Unit test on real data (french auto insurance)
   - Refactor unit tests as there is a huge amount of duplicated code
   - Unit tests on larger matrices
   - Unit tests on csc matrices
   - Unit test for same results with `use_tabmat=True` and `use_tabmat=True`

src/glum/_transformers.py
MartinStancsicsQC: Which one? 😅 I admit that I'm a bit comma-happy in general but I can't find the problematic one here

MartinStancsicsQC:
Quick performance update. On the French auto insurance dataset (678k observations, 1939 columns when one-hot-encoded) the current implementation finds `fit`s the decollinearizer in under a second. `transform` is even quicker. Correctness is yet to be tested, though.

src/glum/_transformers.py
MatthiasSchmidtblaicherQC: You are right, I misread here :).

MartinStancsicsQC:
I'll close it as per our discussion: it does not really belong in `glum`

PR #655, opened 2023-06-20
Create feature names for non-pandas input, too

MartinStancsicsQC:
Checklist

## Main changes
 - The `feature_names_` attribute is populated for non-pandas input, too. It uses the `X_{i}` format for numerical columns, and (in the case of a `SplitMatrix`) the `C_{i}__{cat}` format for categorical columns. It would be a breaking change for users who relied on the `feature_names_` attribute being unassigned in the case of non-pandas input, but I cannot imagine how that would be usefu ([you never know](https://xkcd.com/1172/), though).
 - A `term_names_` attribute is also assigned during fit. It contains the name of the column in the *original input data* that corresponds to each column of the design matrix. In the case of numerical variables, it is the same as `feature_names_`. In the case of categorical columns, it differs in that it does not include the category at the end.

## Reason
A separate PR for Wald-tests will soon follow, for which it will be useful to have a simple way to refer to columns by names. The same goes for what I call terms here. I am submitting this as a separate PR as it is relatively self-contained and it might be easier to review.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ Changelog
 **New feature**
 
 - Added the complementary log-log (`cloglog`) link function.
+- The `feature_names_` attribute is populated for non-pandas input, as well
+- Added a `term_names_` attribute to the `GeneralizedLinearRegressor` and `GeneralizedLinearRegressorCV` classes. It contains the name of the column in the input data that each coefficient corresponds to.
 
 2.5.2 - 2023-06-02
 ------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,10 +10,10 @@ Changelog
 2.6.0 - UNRELEASED
 ------------------
 
-**New feature**
+**New features**
 
 - Added the complementary log-log (`cloglog`) link function.
-- The `feature_names_` attribute is populated for non-pandas input, as well
+- The `feature_names_` attribute is populated for non-pandas input, as well.
 - Added a `term_names_` attribute to the `GeneralizedLinearRegressor` and `GeneralizedLinearRegressorCV` classes. It contains the name of the column in the input data that each coefficient corresponds to.
 
 2.5.2 - 2023-06-02
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -219,9 +219,9 @@ def _name_categorical_variables(
         f"{column_name}__{category}" for category in categories[int(drop_first) :]
     ]
     if len(new_names) == 0:
-        warnings.warn(
+        raise ValueError(
             f"Categorical column: {column_name}, contains only one category. "
-            + "This will be dropped from the feature matrix."
+            + "This should be dropped from the feature matrix."
         )
     return new_names
 
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -1866,6 +1866,54 @@ class GeneralizedLinearRegressorBase(BaseEstimator, RegressorMixin):
 
         return X, y, sample_weight, offset, weights_sum, P1, P2
 
+    def _get_feature_names(self, X: ArrayLike):
+        self.feature_names_ = []
+        self.column_names_ = []
+
+        if isinstance(X, pd.DataFrame):
+            for column, dtype in zip(X.columns, X.dtypes):
+                if pd.api.types.is_categorical_dtype(dtype):
+                    for name in _name_categorical_variables(
+                        dtype.categories, column, self.drop_first
+                    ):
+                        self.feature_names_.append(name)
+                        self.column_names_.append(column)
+                else:
+                    self.feature_names_.append(column)
+                    self.column_names_.append(column)
+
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -13,6 +13,8 @@ Changelog
 **New feature**
 
 - Added the complementary log-log (`cloglog`) link function.
+- The `feature_names_` attribute is populated for non-pandas input, as well
```
MatthiasSchmidtblaicherQC, line 16:
```suggestion
- The `feature_names_` attribute is populated for non-pandas input, as well.
```

src/glum/_glm.py
```diff
@@ -1873,8 +1863,70 @@ def _expand_categorical_penalties(penalty, X, drop_first):
         elif isinstance(X, np.ndarray):
             X = tm.DenseMatrix(X)
 
+        if not hasattr(self, "feature_names_") or not hasattr(self, "term_names_"):
+            self._get_feature_names(X)
```
MatthiasSchmidtblaicherQC, line 1867: Given that feature names are always created, why do they need to be set separately for pd.DataFrames and other objects? I.e., why not move `self._get_feature_names(X)` in L. 1749 out of the condition to line 1746, say?

MartinStancsicsQC: That was my first attempt, but the problem is that at that point it is not certain that `X` is a nice, 2-dimensional array-like object. In particular, `sklearn.utils.estimator_check.check_estimator` is part of the test suite, and it feeds our estimator all kinds of weird objects that `sklearn.utils.check_array` is supposed to handle.

CHANGELOG.rst
MatthiasSchmidtblaicherQC: Also: "New feature**s**" in L.13.

src/glum/_glm.py
```diff
@@ -1873,8 +1863,70 @@ def _expand_categorical_penalties(penalty, X, drop_first):
...
+
         return X, y, sample_weight, offset, weights_sum, P1, P2
 
+    def _get_feature_names(self, X: ArrayLike):
```
jtilly, line 1871: Could this method get a different name, please? I associated "get" with a low overhead getter (i.e. just return an attribute). This method doesn't return anything and it actually sets two fitted attributes. 

I would think it's a bit cleaner if this method just returned the feature names and terms and leave setting them to `_set_up_and_check_fit_args`.

MartinStancsicsQC:
An example notebook demonstrating what this branch is about: [https://gist.github.com/MartinStancsicsQC/eaad758d27cb119336804a3d4170aad1](https://gist.github.com/MartinStancsicsQC/eaad758d27cb119336804a3d4170aad1)

MartinStancsicsQC:
The proposed [tabmat formula interface](https://github.com/Quantco/tabmat/pull/267) has a `categorical_format` attribute that allows parametrizing categorical column names with formula strings (i.e. `"{name}__{category}"` or `"{name}[{category}]"`. I think this feature it would be nice for glum's feature_names, too. Any thoughts?

MatthiasSchmidtblaicherQC:
Re. parametrizing categorical column names: I agree that this is a nice feature, but this should ideally be a separate PR.

src/glum/_glm.py
```diff
@@ -1873,8 +1871,92 @@ def _expand_categorical_penalties(penalty, X, drop_first):
...
+
         return X, y, sample_weight, offset, weights_sum, P1, P2
 
+    def create_feature_and_term_names(
+        self, X: ArrayLike
+    ) -> Tuple[List[str], List[str]]:
+        """
+        Get feature names for the input data.
+
+        Stores them in the ``feature_names_`` and ``column_names_`` attributes.
```
MarcAntoineSchmidtQC, line 1887:
```suggestion
        Stores them in the ``feature_names_`` and ``term_names_`` attributes.
```

MarcAntoineSchmidtQC: Actually, this function doesn't store them. We should revise the docstring.

src/glum/_glm.py
```diff
@@ -1873,8 +1871,92 @@ def _expand_categorical_penalties(penalty, X, drop_first):
...
+                        self.drop_first,
+                        categorical_format=self.categorical_format,
+                    ):
+                        feature_names.append(name)
+                        term_names.append(column)
+                else:
+                    feature_names.append(column)
+                    term_names.append(column)
+
+        elif isinstance(X, tm.StandardizedMatrix):
```
MarcAntoineSchmidtQC, line 1917: We should try to minimize the amount of custom logic depending on the tabmat type. Could you instead create a new method in tabmat to deal with this? I think it would actually be very useful metadata to store in the tabmat object itself.

MarcAntoineSchmidtQC (commented):
My main comment is about the custom handling of different tabmat types. Ideally this is moved into tabmat. Happy to discuss.

MartinStancsicsQC:
It will probably be superseded by [Tabmat PR 278](https://github.com/Quantco/tabmat/pull/278). Closing it for now.

PR #658, opened 2023-06-29
Improve the check for multicollinear columns in `.covariance_matrix()`

MartinStancsicsQC:
Checklist

In the `GeneralizedLinearRegressor.covariance_matrix` method, the condition number (`numpy.linalg.cond`) of `X` is calculated to make sure it is full rank before proceeding. However, this check is only done if the input is a `DenseMatrix`.

This PR proposes to calculate the condition number of `X.T @ X` instead, which is the square of the condition number of `X`. This approach has two main advantages:
 - We can do it for every kind of matrix by leveraging the `.sandwich()` method.
 - It is much faster in the case of tall matrices even when `X` is dense. (In fact, for certain inputs, the majority of the runtime of `covariance_matrix()` was spent on the `numpy.linalg.cond(X)` line before this change.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Changelog
 
 - Added the complementary log-log (`cloglog`) link function.
 
+**Other changes:**
+
+- When computing the covariance matrix, check for ill-conditionedness for all types of input. Furthermore, do it in a more efficient way.
+
 2.5.2 - 2023-06-02
 ------------------
 
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -1490,10 +1490,10 @@ class GeneralizedLinearRegressorBase(BaseEstimator, RegressorMixin):
                 method="pearson",
             )
 
-        if (
-            np.linalg.cond(X.sandwich(np.ones(X.shape[0])))
-            > 1 / sys.float_info.epsilon**2
-        ):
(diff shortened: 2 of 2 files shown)


MartinStancsicsQC:
This PR will hurt performance in the case of **wider-than-tall data**. Should we special-case it and try to come up with something, or just accept it as `glum` is mainly aimed at `k << n` problems?

MatthiasSchmidtblaicherQC:
> This PR will hurt performance in the case of wider-than-tall data.
I don't think that the covariance matrix of coefficient estimates for standard errors and the like will be useful in this case and suspect that computation of the inverse would fail anyway, so I think that we don't need to special case it.

MarcAntoineSchmidtQC (approved):
LGTM

MartinStancsicsQC:
A quick note on why it works. Let us $\kappa, \sigma$ and $\lambda$ denote the condition number, singular value and eigenvalue, respectively. Then

$$ \kappa(X) = \frac{\sigma_{max}{(X)}}{\sigma_{min}{(X)}} = \sqrt{\frac{\lambda_{max}{(X'X)}}{\lambda_{min}{(X'X)}}} = \sqrt{\frac{\sigma_{max}{(X'X)}}{\sigma_{min}{(X'X)}}} = \sqrt{\kappa(X'X)} $$

PR #661, opened 2023-07-07
Add the option to store the covariance matrix to avoid recomputing it

MartinStancsicsQC:
Checklist
* [ ] Added a `CHANGELOG.rst` entry

Add option to store the covariance matrix in the model object when computing it. Furthermore, add a `store_covariance_matrix` parameter to the regressors' constructor, which then automatically estimates and stores the covariance matrix after fitting the model. When a covariance matrix is stored, inference methods can be called without any arguments.

The purpose of this PR is twofold: (1) it makes statistical inference slightly more ergonomic, and (2) it avoids the possibly expensive step of recomputing the covariance matrix when multiple statistical tests are performed. The default is still *not* to store the covariance matrix, and changes are backwards-compatible.

At the moment, only the `std_errors` and `covariance_matrix` methods are affected. A subsequent PR will introduce more statistical inference methods (Wald-tests and p-values), which will also be able to use the stored covariance matrix.

PR #668, opened 2023-08-02
Add methods for performing Wald-tests

MartinStancsicsQC:
Checklist

## Summary

This PR adds some support for doing inference on estimated GLMs. Wald-tests can be performed based on a restriction matrix and corresponding RHS vector (lower-level, flexible), a list of feature names ~~or a list of term names~~.

## Notes

 - ~~This branch is based on the `store-cov-matrix` branch. Therefore, PR #661 should be merged before this one. I am opening this draft PR to get some feedback until that happens.~~
 - ~~Term names is not a concept in glum yet. The idea is based on [Tabmat PR #278](https://github.com/Quantco/tabmat/pull/278) and the closed Glum PR #655. Either these will have to be merged first, or the `GeneralizedLinearRegressorBase.wald_test_term_name` method will have to be removed from this PR until term names are implemented.~~


src/glum/_glm.py
```diff
@@ -1328,6 +1334,293 @@ def predict(
...
+        y=None,
+        mu=None,
+        offset=None,
+        sample_weight=None,
+        dispersion=None,
+        robust=None,
+        clusters: np.ndarray = None,
+        expected_information=None,
+    ) -> WaldTestResult:
+        """Compute the Wald test statistic and p-value for a linear hypotheses.
```
MatthiasSchmidtblaicherQC, line 1440:
```suggestion
        """Compute the Wald test statistic and p-value for a linear hypothesis.
```

src/glum/_glm.py
```diff
@@ -1328,6 +1334,293 @@ def predict(
...
+        y=None,
+        mu=None,
+        offset=None,
+        sample_weight=None,
+        dispersion=None,
+        robust=None,
+        clusters: np.ndarray = None,
+        expected_information=None,
+    ):
+        """Get a table of of the regression coefficeints.
```
MatthiasSchmidtblaicherQC, line 1350:
```suggestion
        """Get a table of of the regression coefficients.
```

src/glum/_glm.py
```diff
@@ -1328,6 +1334,293 @@ def predict(
...
+        Includes coefficient estimates, standard errors, t-values, p-values
```
MatthiasSchmidtblaicherQC, line 1351: Missing blank line.

src/glum/_glm.py
```diff
@@ -1328,6 +1334,293 @@ def predict(
...
+        y=None,
+        mu=None,
+        offset=None,
+        sample_weight=None,
+        dispersion=None,
+        robust=None,
+        clusters: np.ndarray = None,
+        expected_information=None,
+    ) -> WaldTestResult:
+        """Compute the Wald test statistic and p-value for a linear hypotheses.
```
MatthiasSchmidtblaicherQC, line 1542: see above

src/glum/_glm.py
```diff
@@ -1328,6 +1334,293 @@ def predict(
...
+            computed if clusters is not None.
+        expected_information : boolean, optional, default=None
+            Whether to use the expected or observed information matrix.
+            Only relevant when computing robust std-errors.
+            If not specified, the model's ``expected_information`` attribute is used.
+
+        Returns
+        -------
+        WaldTestResult
+            NamedTuple with test statistic, p-value and degrees of freedom.
```
MatthiasSchmidtblaicherQC, line 1481:
```suggestion
            NamedTuple with test statistic, p-value, and degrees of freedom.
```

src/glum/_glm.py
```diff
@@ -1328,6 +1334,293 @@ def predict(
...
+
+        beta = beta[:, np.newaxis]
+        r = r[:, np.newaxis]
+        Q = R.shape[0]
+
+        Rb_r = R @ beta - r  # R \beta - r
+        RVR = R @ covariance_matrix @ R.T  # R V R^T
+
+        # We want to calculate Rb_r^T (RVR)^{-1} Rb_r.
+        # We can do it in a more numerically stable way by using `scipy.linalg.solve`:
```
MatthiasSchmidtblaicherQC, line 1519: neat

tests/glm/test_glm.py
```diff
@@ -1988,6 +1988,208 @@ def test_inputtype_std_errors(regression_data, categorical, split, fit_intercept
...
+    mdl_sm = sm.GLM(endog=y, exog=X, family=sm.families.Gaussian())
+
+    # Use the same covariance matrix for both so that we can use tighter tolerances
+    our_results = mdl.wald_test_matrix(R, r)
+    fit_sm = mdl_sm.fit()
+    sm_results = fit_sm.wald_test((R, r), cov_p=mdl.covariance_matrix(), scalar=False)
+
+    np.testing.assert_allclose(
+        our_results.test_statistic, sm_results.statistic, rtol=1e-8
+    )
```
MatthiasSchmidtblaicherQC, line 2144: Do we have any good rule of ascertaining how big the expected difference with statsmodels due to numerical imprecisions should be? I.e. are we sure that `1e-3` including recomputation of covariance matrix and `1e-8` without should be expected?

One way of checking this would be by comparing deviations between statsmodels and other packages (e.g., GLM in R) for the same problem and checking that they would be of the same order (no need for a test, could just be a one-off analysis).

MatthiasSchmidtblaicherQC (commented):
Looks good to me. I have some small stylistic points and a suggestion for a check that we could do.

tests/glm/test_glm.py
MartinStancsicsQC: Good idea, I'll check against R

src/glum/_glm.py
```diff
@@ -1328,6 +1334,293 @@ def predict(
...
+        clusters: np.ndarray = None,
+        expected_information=None,
+    ):
+        """Get a table of of the regression coefficeints.
+        Includes coefficient estimates, standard errors, t-values, p-values
+        and confidence intervals.
+
+        Parameters
+        ----------
+        significance_level : float, optional (default=0.05)
```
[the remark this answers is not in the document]
MartinStancsicsQC: removed

src/glum/_glm.py
```diff
@@ -1328,6 +1334,293 @@ def predict(
...
+            Training data. Can be omitted if a covariance matrix has already
+            been computed.
+        y : array-like, shape (n_samples,), optional
+            Target values. Can be omitted if a covariance matrix has already
+            been computed.
+        mu : array-like, optional, default=None
+            Array with predictions. Estimated if absent.
+        offset : array-like, optional, default=None
+            Array with additive offsets.
+        sample_weight : array-like, shape (n_samples,), optional (default=None)
```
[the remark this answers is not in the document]
MartinStancsicsQC: This line was copied from the docstring of `covariance_matrix`, where it has parentheses. Should we aim for docstring consistency within a function, or consistency for the same argument across functions?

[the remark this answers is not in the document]
MatthiasSchmidtblaicherQC: Ouch. My preference would be to write it here in the style that we want to follow, I guess `sample_weight : array-like, shape (n_samples,), optional, default=None` and to clean up the other docstrings in a separate future PR.

MatthiasSchmidtblaicherQC (approved):

src/glum/_glm.py
```diff
@@ -1328,6 +1334,295 @@ def predict(
...
+        # We can do it in a more numerically stable way by using `scipy.linalg.solve`:
+        try:
+            test_stat = float(Rb_r.T @ linalg.solve(RVR, Rb_r))
+        except linalg.LinAlgError as err:
+            raise linalg.LinAlgError("The restriction matrix is not full rank") from err
+        p_value = 1 - stats.chi2.cdf(test_stat, Q)
+
+        return WaldTestResult(test_stat, p_value, Q)
+
+    def wald_test_feature_name(
```
MarcAntoineSchmidtQC, line 1530: not a huge fan of the method name.
Here's a suggestion: combine the two methods into one called `wald_test`.
r and values have the same meaning. R and features could share the same argument. If the argument is a list of string, you know that the user wants to use feature name. If it's a numpy array of numeric type, you know the user want the "matrix" form.

To make the change, you could make the two methods private (start with an underscore), and have `wald_test` being a small function to check which one the user wants.

I'm also okay if you keep the current setup, but would recommend you think of a better method name for this one.

MarcAntoineSchmidtQC: I just saw that you are calling `wald_test_matrix` from this method. Maybe you could simply rename this method, and check at the start if the first argument is a numeric array, then directly call `_wald_test_matrix`.

MarcAntoineSchmidtQC (approved):
Looks really good.
Please look at my suggestion for renaming the wald test methods. Otherwise, I think this can be merged.
🎉

src/glum/_glm.py
MartinStancsicsQC: Good point, I'm not a big fan of the current names, either. Dispatching based on the type of `R` would work for the current two functions, but in `v3` I'm planning to add formula-based and term-name-based Wald-tests, and `R` would be a string / list of strings in those, too.

What do you think of an in-between solution? Have one public Wald-test function and make the current ones private, and the publc function would accept `R_matrix`, `features` (and later on `terms` and `formula`) parameters. Dispatch would be based on which of these is provided. We would avoid the weird names, but choosing how to test would remain a bit more explicit?

MarcAntoineSchmidtQC: Yeah, that's a very good solution!

MartinStancsicsQC:
Thank you! The methods have been renamed. Should it go into `main` or `glum-v3`?

src/glum/_glm.py
```diff
@@ -1328,6 +1334,401 @@ def predict(
...
+        expected_information=None,
+    ):
+        """Get a table of of the regression coefficients.
+
+        Includes coefficient estimates, standard errors, t-values, p-values
+        and confidence intervals.
+
+        Parameters
+        ----------
+        significance_level : float, optional, default=0.05
```
MatthiasSchmidtblaicherQC, line 1357: What about instead having an argument `confidence_level` with default value 0.95? This seems to be slightly more common language (e.g., the [Wikipedia article](https://en.wikipedia.org/wiki/Confidence_interval) only mentions confidence levels) "Significance level" is used mostly in the context of hypothesis testing.

src/glum/_glm.py
```diff
@@ -1328,6 +1334,401 @@ def predict(
...
+            Array with additive offsets.
+        sample_weight : array-like, shape (n_samples,), optional, default=None
+            Individual weights for each sample.
+        dispersion : float, optional, default=None
+            The dispersion parameter. Estimated if absent.
+        robust : boolean, optional, default=None
+            Whether to compute robust standard errors instead of normal ones.
+            If not specified, the model's ``robust`` attribute is used.
+        clusters : array-like, optional, default=None
+            Array with clusters membership. Clustered standard errors are
```
MatthiasSchmidtblaicherQC, line 1377: Shouldn't it be **cluster** membership (singular)?

src/glum/_glm.py
```diff
@@ -1328,6 +1334,401 @@ def predict(
...
+            computed if clusters is not None.
+        expected_information : boolean, optional, default=None
+            Whether to use the expected or observed information matrix.
+            Only relevant when computing robust std-errors.
```
MatthiasSchmidtblaicherQC, line 1485:
```suggestion
            Only relevant when computing robust standard errors.
```

src/glum/_glm.py
```diff
@@ -1328,6 +1334,401 @@ def predict(
...
+        offset=None,
+        sample_weight=None,
+        dispersion=None,
+        robust=None,
+        clusters: np.ndarray = None,
+        expected_information=None,
+    ) -> WaldTestResult:
+        """Compute the Wald test statistic and p-value for a linear hypothesis.
+
+        The hypothesis tested is ``R @ coef_ = r``. The test statistic follows
```
MatthiasSchmidtblaicherQC, line 1549: Should mention that: "Under the null hyptothesis, the test statistics follows a chi-squared distribution..." (not under the alternative or at least not against all alternatives).

src/glum/_glm.py
```diff
@@ -1328,6 +1334,401 @@ def predict(
...
+            Array with additive offsets.
+        sample_weight : array-like, shape (n_samples,), optional, default=None
+            Individual weights for each sample.
+        dispersion : float, optional, default=None
+            The dispersion parameter. Estimated if absent.
+        robust : boolean, optional, default=None
+            Whether to compute robust standard errors instead of normal ones.
+            If not specified, the model's ``robust`` attribute is used.
+        clusters : array-like, optional, default=None
+            Array with clusters membership. Clustered standard errors are
```
MatthiasSchmidtblaicherQC, line 1579: see above

src/glum/_glm.py
```diff
@@ -1328,6 +1334,401 @@ def predict(
...
+            Only relevant when computing robust std-errors.
+            If not specified, the model's ``expected_information`` attribute is used.
+
+        Returns
+        -------
+        WaldTestResult
+            NamedTuple with test statistic, p-value, and degrees of freedom.
+        """
+
+        if isinstance(features, str):
```
MatthiasSchmidtblaicherQC, line 1693: Do we need a check for whether features are None or an empty iterable?

tests/glm/test_glm.py
```diff
@@ -1988,6 +1988,277 @@ def test_inputtype_std_errors(regression_data, categorical, split, fit_intercept
...
+        ),
+        pytest.param(
+            ["intercept", "col_9"],
+            np.array([[1] + [0] * 10, [0] * 10 + [1]]),
+            [1, 2],
+            id="intercept",
+        ),
+    ],
+)
+def test_wald_test_feature_name(regression_data, names, R, r):
```
MatthiasSchmidtblaicherQC, line 2198:
```suggestion
def test_wald_test_feature_names(regression_data, names, R, r):
```

tests/glm/test_glm.py
```diff
@@ -1988,6 +1988,277 @@ def test_inputtype_std_errors(regression_data, categorical, split, fit_intercept
...
+            id="rhs_not_zero",
+        ),
+        pytest.param(
+            ["intercept", "col_9"],
+            [1, 2],
+            id="intercept",
+        ),
+    ],
+)
+def test_wald_test_feature_name_public(regression_data, names, r):
```
MatthiasSchmidtblaicherQC, line 2239:
```suggestion
def test_wald_test_feature_names_public(regression_data, names, r):
```

MatthiasSchmidtblaicherQC (changes requested):

src/glum/_glm.py
MartinStancsicsQC: Hmm, good point. I'll test what happens with those inputs

tests/glm/test_glm.py
```diff
@@ -1988,6 +1988,277 @@ def test_inputtype_std_errors(regression_data, categorical, split, fit_intercept
...
+    np.testing.assert_allclose(our_table["se"], fit_sm.bse, rtol=1e-8)
+    np.testing.assert_allclose(our_table["t_value"], fit_sm.tvalues, rtol=1e-8)
+    np.testing.assert_allclose(our_table["p_value"], fit_sm.pvalues, atol=1e-8)
+    np.testing.assert_allclose(
+        our_table[["ci_lower", "ci_upper"]], fit_sm.conf_int(), rtol=1e-8
+    )
+
+
+@pytest.mark.parametrize("fit_intercept", [True, False])
+@pytest.mark.parametrize("family", ["poisson", "normal", "binomial"])
```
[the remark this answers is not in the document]
MartinStancsicsQC: Not really. I just did not want to spend too much test/CI resources on this, because they can only differ through the estimated covariance matrix, which should be covered in more specific tests.

src/glum/_glm.py
MartinStancsicsQC: Calling the function with `None` returns a `TypeError: object of type 'NoneType' has no len()` exception, which is fine for non-end-users. End users using the public `wald_test` function should never encounter a case when this private function is called with `names=None`.

Calling it with an empty list returns a test-statistic of 0 (correct), and a p-value of `nan` (should be `1.0`, but I'm not sure it's worth special-casing it. I think these should make it obvious to the user what happened.

MartinStancsicsQC:
Note: windows conda-build also fails on [main](https://github.com/Quantco/glum/actions/runs/5927871847/job/16104923901). (was fine in the morning...)

MartinStancsicsQC:
Related mamba issue: https://github.com/mamba-org/mamba/issues/2770

MartinStancsicsQC:
I don't think so. I'm not even sure where that pin comes from.

MartinStancsicsQC:
@jonashaag I don't think it's about the libmamba version. The CI for the same commit was [green this morning](https://github.com/Quantco/glum/actions/runs/5926807015) but [red when I rerun it this afternoon.](https://github.com/Quantco/glum/actions/runs/5927871847). Based on the logs, the difference between the `build` environment is that the passing one uses `libsolv` 0.7.24 **build 1**, while the failing one uses `libsolv` 0.7.24 **build 2**. (there is also a difference in the patch version of `wheel` but I don't think that's the issue.) Is it possible that it is a `libsolv` problem?

jonashaag:
@pavelzw it looks like you made the libsolv change, can you have a look please?

MartinStancsicsQC:
Never mind, libsolv build 2 [was indeed broken](https://github.com/conda-forge/admin-requests/pull/800) and build 3 [seem to have solved our issues](https://github.com/Quantco/glum/actions/runs/5939803442/job/16110515136)

src/glum/_glm.py
MatthiasSchmidtblaicherQC: Ok, thanks for checking.

MatthiasSchmidtblaicherQC (approved):

Issue #669, opened 2023-08-02
`covariance_matrix()` can overwrite feature names


MartinStancsicsQC:
## Description

When a the `covariance_matrix` method of a fitted `GeneralizedLinearRegressor` object is called with inputs that have different colum  names than the one the model was trained on, the column names are silently overwritten.

## Example

```python
>>> import pandas as pd
>>> import glum

>>> df = pd.DataFrame({
...     "x1": [1.0, 2.0, 3.0],
...     "x2": [2.0, 3.0, 4.0],
...     "y": [3.0, 2.0, 0.0]
... })

>>> model = glum.GeneralizedLinearRegressor().fit(df[["x1"]], df["y"])
>>> model.feature_names_
Index(['x1'], dtype='object')

>>> _ = model.covariance_matrix(df[["x2"]], df["y"])
>>> model.feature_names_
Index(['x2'], dtype='object')
```

## Proposed solution

The reason for this bug is that `covariance_matrix` uses the `_set_up_and_check_fit_args` method, which presumes it is called during fit, and sets up various model attributes in addition to returning some inputs. (It also does some work that is unnecessary for `covariance_matrix`, such as expanding categorical penalties. As a quick fix, `check_X_y_tabmat_compliant` could be used instead, like in `predict`.

PR #671, opened 2023-08-03
Fix covariance matrix mutating feature names

MartinStancsicsQC:
Checklist

This PR implements a minimal fix for #669. It replaces the (non-pure) `_set_up_and_check_fit_args` method call in `covariance_matrix()` with the side-effectless `check_X_y_tabmat_compliant` function plus a few more lines of code.

The fix does work, but I believe that the underlying issue is a bit deeper, and a proper solution would require some refactoring. Basically, it is not totally clear (at least to me) where the line lies between `_set_up_and_check_fit_args` and `check_X_y_tabmat_compliant`. For example, converting to `MatrixBase` and handling integer arrays is done in the former, while I feel it should belong to the latter.

It would be nice to have a method/function (could be `check_X_y_tabmat_compliant`), after which we can be sure that `X` is a nice, checked `MatrixBase` object, but which does not have side effects. It could also take care of aligning categorical features (I believe that the category alignment step was omitted from `covariance_matrix` before.), or, later on, materializing formuals.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Changelog
 
 - Added the complementary log-log (`cloglog`) link function.
 
+**Bug fix**
+
+- Fixed :meth:`~glum.GeneralizedLinearRegressorBase.covariance_matrix` mutating feature names when called with a data frame. See `here <https://github.com/Quantco/glum/issues/669>`_.
+
 **Other changes:**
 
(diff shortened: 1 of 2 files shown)


lbittarello:
Change log? :)

MartinStancsicsQC:
Oops, added :) Do you think we should merge this quick fix, or go for refactoring `check_X_y_tabmat_compliant` and `_set_up_and_check_fit_args`? (Or do both in this order?)

PR #678, opened 2023-08-14
Column names  à la Tabmat #278

MartinStancsicsQC:
Checklist

Delegate column name creation to tabmat when calling `from_pandas`. As an additional perk, non-pandas inputs will also have auto-generated column (and term) names, making performing Wald-tests much simpler.


tests/glm/test_glm.py
```diff
@@ -2090,15 +2199,15 @@ def test_drop_first_allows_alpha_equals_0():
         regressor.fit(X, y)
 
 
-def test_error_on_distinct_categorical_column():
+@pytest.mark.xfail
```
MarcAntoineSchmidtQC, line 2202: What exactly is this supposed to test? Having a xfail mark expects a failure, but there are 3 different asserts in this test. This means that we don't know which one failed. Are they all supposed to fail?

MarcAntoineSchmidtQC (approved):
Looks good. Only one comment for a test.

tests/glm/test_glm.py
MartinStancsicsQC: Good catch. I think we should get rid of that test. It used to be about raising an error if a categorical column has only one category (as that would have complicated naming the column). Now however, `tabmat` handles column names properly even in that case, and I don't think singleton categories should raise an error (maybe a warning?), as they are correctly handled by not including any column for them in the matrix.

PR #684, opened 2023-08-17
Support for missing values in categorical columns

MartinStancsicsQC:
Checklist


src/glum/_glm.py
```diff
@@ -2251,12 +2266,23 @@ class GeneralizedLinearRegressor(GeneralizedLinearRegressorBase):
         If true, then the expected information matrix is computed by default.
         Only relevant when computing robust standard errors.
 
-    categorical_features : str, optional (default = "{name}[{category}]")
+    categorical_format : str, optional (default = "{name}[{category}]")
```
MatthiasSchmidtblaicherQC, line 2269:
```suggestion
    categorical_format : str, optional, default='{name}[{category}]'
```

src/glum/_glm.py
```diff
@@ -2251,12 +2266,23 @@ class GeneralizedLinearRegressor(GeneralizedLinearRegressorBase):
...
         Format string for categorical features. The format string should
         contain the placeholder ``{name}`` for the feature name and
         ``{category}`` for the category name. Only used if ``X`` is a pandas
         DataFrame.
 
+    cat_missing_method: str {'fail'|'zero'|'convert'}, default 'fail'
```
MatthiasSchmidtblaicherQC, line 2275:
```suggestion
    cat_missing_method: str {'fail'|'zero'|'convert'}, default='fail'
```

MatthiasSchmidtblaicherQC: We should unify the notation for default parameters in a different PR (not pertaining to your changes here). The notation with defaults in parenthesis does not seem to be an [officially recognized style](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) though.

src/glum/_glm_cv.py
```diff
@@ -281,11 +281,22 @@ class GeneralizedLinearRegressorCV(GeneralizedLinearRegressorBase):
...
         Only relevant when computing robust standard errors.
 
-    categorical_features : str, optional (default = "{name}[{category}]")
+    categorical_format : str, optional (default = "{name}[{category}]")
         Format string for categorical features. The format string should
         contain the placeholder ``{name}`` for the feature name and
         ``{category}`` for the category name. Only used if ``X`` is a pandas
         DataFrame.
+
+    cat_missing_method: str {'fail'|'zero'|'convert'}, default 'fail'
```
MatthiasSchmidtblaicherQC, line 290: see above

src/glum/_util.py
```diff
@@ -53,6 +53,40 @@ def _align_df_categories(df, dtypes) -> pd.DataFrame:
...
+) -> pd.DataFrame:
+    if not isinstance(df, pd.DataFrame):
+        raise TypeError(f"Expected `pandas.DataFrame'; got {type(df)}.")
+
+    changed_dtypes = {}
+
+    categorical_dtypes = [
+        column
+        for column, dtype in dtypes.items()
+        if pd.api.types.is_categorical_dtype(dtype) and (column in df)
```
MatthiasSchmidtblaicherQC, line 71: Are there cases in which it makes sense to fit or predict on a df that does not have all features in `self.feature_dtypes_,`? If we can't think of any, we should raise if the column cannot be found in df. We can always relax this later in case that a use case comes up.

src/glum/_util.py
```diff
@@ -53,6 +53,40 @@ def _align_df_categories(df, dtypes) -> pd.DataFrame:
...
+    ]
+
+    for column in categorical_dtypes:
+        if (
+            categorical_format.format(name=column, category=cat_missing_name)
+            in feature_names
+        ):
+            _logger.info(f"Adding missing category {cat_missing_name} to {column}.")
+            changed_dtypes[column] = df[column].cat.add_categories(cat_missing_name)
```
MatthiasSchmidtblaicherQC, line 80: This will raise in case that `cat_missing_name` is already a category, right? That is intended?

tests/glm/test_utils.py
```diff
@@ -96,3 +96,52 @@ def test_align_df_categories_missing_columns(df):
 def test_align_df_categories_not_df():
     with pytest.raises(TypeError):
         _align_df_categories(np.array([[0], [1]]), {"x0": np.float64})
+
+
+@pytest.fixture()
+def df_na():
```
MatthiasSchmidtblaicherQC, line 102: Would be good to have tests for the different cases mentioned above: a data frame with a categorical column that already contains `"(M)"` and a data frame that does not contain all dtypes.

MatthiasSchmidtblaicherQC (changes requested):
Looks generally good to me, but it looks like we should think more carefully about edge cases before merging.

src/glum/_util.py
MartinStancsicsQC: Good point. I was basing it off `_align_df_categories` which has the same logic. It might make sense to raise in both functions, though. Or, maybe, it could be part of a more general check (`df` has the expected columns in the expected order) in a separate function.

src/glum/_util.py
MartinStancsicsQC: Yes that's intended. We might want to have a nicer error message here, though?

src/glum/_glm.py
MartinStancsicsQC: Agreed.

src/glum/_glm.py
```diff
@@ -837,11 +841,20 @@ def _convert_from_pandas(self, df: pd.DataFrame) -> tm.MatrixBase:
         """Convert a pandas data frame to a tabmat matrix."""
         if hasattr(self, "feature_dtypes_"):
             df = _align_df_categories(df, self.feature_dtypes_)
+            if self.cat_missing_method == "convert":
```
MarcAntoineSchmidtQC, line 844: Here is the logic for when `cat_missing_method` is "convert", but I can't find the logic for handling fail vs zero.

MarcAntoineSchmidtQC: Is it implemented in tabmat?

MartinStancsicsQC: Yes, it is implemented in tabmat. Creating a new category for the missings is also implemented in tabmat (e.g. the `fit` method does not need to have this logic), but we have to do it manually in predict-type functions to make sure that we are doing it the same way as at fitting time (e.g. if a categorical column has missing values in the training data, but no missing values in the data used for prediction, we still need to create the missing category for it in the latter).

src/glum/_util.py
MartinStancsicsQC: @MatthiasSchmidtblaicherQC let me know if you are okay with doing it in a future PR (as the idea is somewhat more general than categorical missings IMO)

MatthiasSchmidtblaicherQC: Yeah sounds good. In case you don't make a PR until end of next week, can you open an issue?

MatthiasSchmidtblaicherQC (approved):

MarcAntoineSchmidtQC (approved):

PR #704, opened 2023-10-09
Drop Python<3.9 according to NEP29

xhochy:
Checklist


jtilly (approved):
Maybe also remove https://github.com/Quantco/glum/blob/eb3d08933809378f6179590081ac2d5c0045b5b2/setup.py#L69?

xhochy:
> Maybe also remove
> 
> https://github.com/Quantco/glum/blob/eb3d08933809378f6179590081ac2d5c0045b5b2/setup.py#L69
> 
> ?

Done.

PR #715, opened 2023-10-19
Let users skip the covariance in `coef_table`

lbittarello:
Inference doesn't always make sense (e.g., if the model is regularised), but we may still want to look at the coefficients.


MarcAntoineSchmidtQC (approved):
Looks good.

PR #718, opened 2023-10-25
Update mirror for nightlies

jtilly:
[scientific-python-nightly-wheels](https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/) is a lot more up to date than [scipy-wheels-nightly](https://pypi.anaconda.org/scipy-wheels-nightly/simple/scikit-learn/).

CI failure is because of https://github.com/Quantco/glum/issues/714.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -36,7 +36,7 @@ jobs:
           echo "Install compilation dependencies"
           micromamba install -y c-compiler cxx-compiler cython jemalloc-local libgomp mako xsimd
 
-          PRE_WHEELS="https://pypi.anaconda.org/scipy-wheels-nightly/simple"
+          PRE_WHEELS="https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/"
           for pkg in numpy pandas scikit-learn scipy; do
             echo "Installing $pkg nightly"
             micromamba remove -y --force $pkg
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -2,6 +2,7 @@ name: Daily runs
 on:
   schedule:
     - cron: '0 5 * * *'
+  workflow_dispatch:
   push:
     paths:
     - '.github/workflows/daily.yml'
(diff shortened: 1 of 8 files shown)


lbittarello (approved):

MarcAntoineSchmidtQC:
With the last commit we are now able to build the environment. The tests are failing because the new channel points to numpy 2.0. From what I can see, there's a simple fix (switch from `np.Inf` to `np.inf`), and there's one that I'm not sure yet (one about [NEP50](https://numpy.org/neps/nep-0050-scalar-promotion.html))

.github/workflows/daily.yml
```diff
@@ -34,17 +34,17 @@ jobs:
         run: |
           # needed for tabmat
           echo "Install compilation dependencies"
-          micromamba install -y c-compiler cxx-compiler cython jemalloc-local libgomp mako xsimd
+          micromamba install -y c-compiler cxx-compiler 'cython!=3.0.4' jemalloc-local libgomp mako xsimd
 
-          PRE_WHEELS="https://pypi.anaconda.org/scipy-wheels-nightly/simple"
+          PRE_WHEELS="https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/"
```
MarcAntoineSchmidtQC, line 39: Adding a review comment so that we do not merge this right now. We need to fix the errors caused by the newer versions.

MarcAntoineSchmidtQC (changes requested):

MarcAntoineSchmidtQC:
Reading a bit on numpy 2.0. I'm worried that the switch to [scientific-python-nightly-wheels](https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/) will be problematic. The fact that it's more up to date also has a downside: the probability that those packages are not yet working together is much higher.

Currently, Pandas does not work with numpy 2.0: https://github.com/pandas-dev/pandas/issues/55519.

jtilly:
Yep, got it. But that just means we shouldn't test against NumPy 2.0. I think it doesn't justify using an outdated mirror that won't get updated any longer 😉

MarcAntoineSchmidtQC (approved):

PR #720, opened 2023-10-30
Fix nightly 2.0

MarcAntoineSchmidtQC:
Fixing glum to support numpy 2.0. Here is a summary of the changes:

- directly support numpy 2.0:  e.g. `np.Inf` -> `np.inf`
- also use a nightly install of pyarrow (since the stable version doesn't support numpy 2.0)
- some weird changes that should be monitored: statsmodels results are returned as a 2d array while we are using 1d array. The test used to work but now fails. Seems to now be okay with both numpy 2.0 and 1.26, so the likely culprit is probably that before numpy 2.0, [[a]] == [a].

Checklist

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -2,6 +2,7 @@ name: Daily runs
 on:
   schedule:
     - cron: '0 5 * * *'
+  workflow_dispatch:
   push:
     paths:
     - '.github/workflows/daily.yml'
@@ -37,8 +38,7 @@ jobs:
           micromamba install -y c-compiler cxx-compiler 'cython!=3.0.4' jemalloc-local libgomp mako xsimd
 
           PRE_WHEELS="https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/"
-          # FIXME: re-add NumPy when pandas supports NumPy>=2
-          for pkg in pandas scikit-learn scipy; do
+          for pkg in numpy pandas scikit-learn scipy; do
             echo "Installing $pkg nightly"
             micromamba remove -y --force $pkg
             pip install --pre --no-deps --only-binary :all: --upgrade --timeout=60 -i $PRE_WHEELS $pkg
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -45,7 +45,7 @@ jobs:
           done
           echo Install pyarrow nightly
           micromamba remove -y --force pyarrow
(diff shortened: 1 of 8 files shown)


.github/workflows/daily.yml
```diff
@@ -37,12 +38,14 @@ jobs:
...
-          # FIXME: re-add NumPy when pandas supports NumPy>=2
-          for pkg in pandas scikit-learn scipy; do
+          for pkg in numpy pandas scikit-learn scipy; do
             echo "Installing $pkg nightly"
             micromamba remove -y --force $pkg
             pip install --pre --no-deps --only-binary :all: --upgrade --timeout=60 -i $PRE_WHEELS $pkg
           done
+          echo Install pyarrow nightly
+          micromamba remove -y --force pyarrow
+          pip install --extra-index-url https://pypi.fury.io/arrow-nightlies/ --prefer-binary --pre --no-deps pyarrow
```
MarcAntoineSchmidtQC, line 48: From: https://arrow.apache.org/docs/developers/python.html#installing-nightly-packages

src/glum/_glm.py
```diff
@@ -929,7 +929,11 @@ def _make_grid(max_alpha: float) -> np.ndarray:
                     warnings.warn("`min_alpha` is set. Ignoring `min_alpha_ratio`.")
                 min_alpha = self.min_alpha
             return np.logspace(
-                np.log(max_alpha), np.log(min_alpha), self.n_alphas, base=np.e
+                np.log(max_alpha),
+                np.log(min_alpha),
+                self.n_alphas,
+                base=np.e,
+                dtype=X.dtype,
```
MarcAntoineSchmidtQC, line 936: Added the dtype here. It was causing issues with 32-bit runs because the alphas were set to be 64 bits, and this would upcast downstream variables when multiplying, causing typing issues with our C++ code.

src/glum/_glm.py
```diff
@@ -1631,7 +1635,7 @@ def _wald_test_matrix(
         # We want to calculate Rb_r^T (RVR)^{-1} Rb_r.
         # We can do it in a more numerically stable way by using `scipy.linalg.solve`:
         try:
-            test_stat = float(Rb_r.T @ linalg.solve(RVR, Rb_r))
```
MarcAntoineSchmidtQC, line 1634: not allowed anymore with numpy 2.0

jtilly (approved):

Issue #723, opened 2023-10-31
GLM tests of scikit-learn


lorentzenchr:
Scikit-learn has some very strict tests for GLMs in https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/linear_model/_glm/tests/test_glm.py. I modified the file to test `glum.GeneralizedLinearRegressor` instead, see https://gist.github.com/lorentzenchr/2e319bcfd4aadfbea64c6330e5b33521. Running `pytest test_glm.py` results in 76 failed, 212 passed, 104 warnings.

It might be interesting to include those tests in glum.

jtilly:
Thanks a lot! For future reference, these are the failing tests:

<details>

```
FAILED tests_glum.py::test_glm_regression_unpenalized[wide-GeneralizedLinearRegressor(family='binomial')-True-irls-ls] - numpy.linalg.LinAlgError: Matrix is singular.
FAILED tests_glum.py::test_glm_regression_unpenalized[wide-GeneralizedLinearRegressor(family='binomial')-True-irls-cd] - assert 0.690107820640591 == 1.2837501395684472 ± 6.4e-05
FAILED tests_glum.py::test_glm_regression_unpenalized[wide-GeneralizedLinearRegressor(family='binomial')-False-lbfgs] - AssertionError: 
FAILED tests_glum.py::test_glm_regression_unpenalized[wide-GeneralizedLinearRegressor(family='binomial')-False-irls-ls] - numpy.linalg.LinAlgError: Matrix is singular.
FAILED tests_glum.py::test_glm_regression_unpenalized[wide-GeneralizedLinearRegressor(family='binomial')-False-irls-cd] - AssertionError: 
FAILED tests_glum.py::test_glm_regression_unpenalized[wide-GeneralizedLinearRegressor(family='poisson')-True-irls-ls] - numpy.linalg.LinAlgError: Matrix is singular.
FAILED tests_glum.py::test_glm_regression_unpenalized[wide-GeneralizedLinearRegressor(family='poisson')-True-irls-cd] - assert 0.8533955861703721 == 1.2837501395684472 ± 6.4e-05
FAILED tests_glum.py::test_glm_regression_unpenalized[wide-GeneralizedLinearRegressor(family='poisson')-False-lbfgs] - AssertionError: 
[...]

PR #724, opened 2023-11-02
ENH deal with tiny loss improvements in line search

lorentzenchr:
This PR adds a 2. convergence check based on gradients in the line search. This can deal with tiny loss improvements.

This is taken from https://github.com/scikit-learn/scikit-learn/blob/9621539a9defe86ff55c890d5f2475f42697604f/sklearn/linear_model/_glm/_newton_solver.py#L263.

This might help with some of the tests in #723.

Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Changelog
 
 - Require Python>=3.9 in line with `NEP 29 <https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table>`_
 - Build and test with Python 3.12 in CI.
+- Added line search stopping criterion for tiny loss improvements based on gradient information.
 
 2.6.0 - 2023-09-05
 ------------------
--- a/src/glum/_solvers.py
+++ b/src/glum/_solvers.py
@@ -15,7 +15,7 @@ from ._cd_fast import (
     enet_coordinate_descent_gram,
     identify_active_rows,
 )
-from ._distribution import ExponentialDispersionModel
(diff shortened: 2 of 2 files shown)


MarcAntoineSchmidtQC (approved):
I'm good with this change. Does not have any negative performance impact on our current benchmarks, and the logic seems to make sense. The new conditional branch is rarely used in our current golden master tests, but used quite often in the `test_glm.py` tests that @lorentzenchr provided.

Two small requests before merging: 
- Is it possible to provide a source for the logic behind the tiny loss branch?
- Can you add an entry to the changelog in the "Unreleased" section.

lorentzenchr:
> Is it possible to provide a source for the logic behind the tiny loss branch?

I developed this in https://github.com/scikit-learn/scikit-learn/pull/24637, trying to satisfy my own (very strict) tests. I can't provide any literature on this.
The basic insight is as follows:
- In general, the first order condition alias score equation is a better condition for checking convergence than loss improvements or step sizes (change in coefficients) because we know the absolute scale: it should be zero.
- Empirically, if we are already close to the minimum and seeking for a high precision solution, the floating point precision of the loss might not be enough while the gradient has not yet achieved, say 1e-12 or lower. Therefore, the gradient has more information in such a setting.
- The line search may then reject a (Newton) step that brings the gradient closer to zero, but that provides no improvement of the loss (but no worsening either up to machine precision).

MarcAntoineSchmidtQC:
Thanks @lorentzenchr!

PR #726, opened 2023-11-02
Fix cython performance regression

### LINKED ISSUE #721: Performance of Tweedie log-likelihood
I know, this is not the primary focus of glum. But I thought it might be interesting:

```python
import numpy as np
from scipy.special import wright_bessel
from glum._distribution import TweedieDistribution

def tweedie_1(y, mu, p, phi):
    # We omit handling y==0 for brevity.
    theta = mu ** (1 - p) / (1 - p)
    kappa = mu ** (2 - p) / (2 - p)
    alpha = (2 - p) / (1 - p)
    ll_obs = (y * theta - kappa) / phi

    x = ((p - 1) * phi / y) ** alpha
    x /=  (2 - p) * phi
    wb = wright_bessel(-alpha, 0, x)
    ll_obs += np.log(wb / y)
    
    return np.sum(ll_obs)

def tweedie_2(y, mu, p, phi):
    td = TweedieDistribution(power=p)
    return td.log_likelihood(
        y=y,
        mu=mu,
        dispersion=phi,
    )

rng = np.random.default_rng(42)
n = 10_000
y = rng.gamma(2, scale=1.5, size=n)
mu = rng.gamma(2, scale=1.5, size=n)
p = 1.5
phi = 1.2

tweedie_1(y, mu, p, phi), tweedie_2(y, mu, p, phi)
# (-26732.28570241986, -26732.285702419867)

%time tweedie_1(y, mu, p, phi)
# CPU times: user 49.4 ms, sys: 1.46 ms, total: 50.9 ms
# Wall time: 50.2 ms

%time tweedie_2(y, mu, p, phi)
# CPU times: user 2.81 s, sys: 10.9 s, total: 13.7 s
# Wall time: 4.04 s
```

So it's almost a factor of 100. The problem with `wright_bessel` is that it is less robust especially for extreme parameters (e.g. internally exp(400) resulting in inf or nan).

MarcAntoineSchmidtQC:
Fixes #721.

With cython 3, a major change in the way functions deal with exceptions happened. Before cython 3, if a function didn't have an explicit handling of exception, exceptions were ignored. With cython 3, the default is for a function to always propagate. This leads to a significant performance hit for us. This PR reverts to cython <3 behavior by using the `legacy_implicit_noexcept` compiler directive.

At the same time, I moved the compiler directives from a comment at the top of the cython files and centralized them in the setup.py file.

Testing with the code snippet in the linked issue gives me back the old performance (around 2ms)

Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,8 +7,12 @@
 Changelog
 =========
 
-2.7.0 - 2023-10-xx
-------------------
+Unreleased
+----------
+
+**Bug fix:**
+
+- Added cython compiler directive legacy_implicit_noexcept = True to fix performance regression with cython 3.
 
 **Other changes:**
 
--- a/setup.py
+++ b/setup.py
@@ -95,6 +95,18 @@ setup(
         glm_benchmarks_run = glum_benchmarks.cli_run:cli_run
         glm_benchmarks_analyze = glum_benchmarks.cli_analyze:cli_analyze
     """,
-    ext_modules=cythonize(ext_modules, annotate=False),
+    ext_modules=cythonize(
+        ext_modules,
+        annotate=False,
+        compiler_directives={
+            "language_level": "3",
+            "boundscheck": False,
+            "wraparound": False,
+            "initializedcheck": False,
+            "nonecheck": False,
+            "cdivision": True,
+            "legacy_implicit_noexcept": True,
+        },
+    ),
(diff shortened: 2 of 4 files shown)


jtilly (approved):
Are we facing the same issue in tabmat?

MarcAntoineSchmidtQC:
Most probably, yes. I will do the same fix there as well.

PR #746, opened 2024-01-09
Fix formula test: consider presence of intercept in full rankness check when constructing the model matrix externally

MatthiasSchmidtblaicherQC:
`test_formula` compares a `GeneralizedLinearRegressor` set up with a `formula` argument to one that receives an externally created model matrix via `formulaic.model_matrix`. The test used to fail for one case because the full rank check in the externally created model matrix did not take into account whether the formula implied an intercept or not.

### DIFF
--- a/tests/glm/test_glm.py
+++ b/tests/glm/test_glm.py
@@ -2957,7 +2957,6 @@ def get_mixed_data():
     )
 
 
-@pytest.mark.skip(reason="Test is not correct")
 @pytest.mark.parametrize(
     "formula",
     [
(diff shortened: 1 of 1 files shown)


jtilly (approved):

MatthiasSchmidtblaicherQC:
One point that came up in testing: specifying a formula without explicitly setting the intercept like `y ~ x1 + x2` will create an intercept when calling `formulaic.model_matrix`, but not in the current `GeneralizedLinearRegressor` (hence addition of "-1" when constructing the model matrix). 

Unless if there was a good reason for this departure from formulaic (?), I would adjust in a separate PR. Alternatively, we could also decide to not allow to set intercept via "+1", "-1", etc. in a formula and, instead, to rely on the `fit_intercept` argument. What do you think (especially @stanmart @MarcAntoineSchmidtQC)?

MarcAntoineSchmidtQC:
I don't mind what takes precedence, but we should make it very clear in the docstring of the estimator.

PR #757, opened 2024-01-30
Do not fail when an estimator misses class members that are new in v3

MatthiasSchmidtblaicherQC:
So that, in v3, we can still use models that were trained on glum < 3.0.

### DIFF
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -906,7 +906,7 @@ class GeneralizedLinearRegressorBase(BaseEstimator, RegressorMixin):
         X = tm.from_pandas(
             df,
             drop_first=self.drop_first,
-            categorical_format=getattr(  # categorical format prior to v3
+            categorical_format=getattr(  # convention prior to v3
                 self, "categorical_format", "{name}__{category}"
             ),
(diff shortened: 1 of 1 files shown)


src/glum/_glm.py
```diff
@@ -2724,7 +2734,8 @@ def _set_up_and_check_fit_args(
 
                 self.feature_dtypes_ = X.dtypes.to_dict()
                 self.has_missing_category_ = {
-                    col: (self.cat_missing_method == "convert") and X[col].isna().any()
+                    col: (getattr(self, "cat_missing_method", "fail") == "convert")
+                    and X[col].isna().any()
                     for col, dtype in self.feature_dtypes_.items()
                     if isinstance(dtype, pd.CategoricalDtype)
                 }
```
MatthiasSchmidtblaicherQC, line 2741: Note that, because `self.has_missing_category_` is created here, we don't need to use `getattr(self, "has_missing_category_", ...)` in 2789 and 2792 anymore.

lbittarello (approved):

stanmart (approved):

Issue #728, opened 2023-11-05
Several warnings and erros in coef_table


lorentzenchr:
With glum version 2.6.0:
```python
from glum import GeneralizedLinearRegressor
import pandas as pd
import numpy as np


X, y = pd.DataFrame({"x": np.arange(3)}), np.ones(3)
glm = GeneralizedLinearRegressor(family="poisson").fit(X, y)
glm.coef_table(X=X, y=y)
```
results in
```
python3.9/site-packages/glum/_glm.py:1909: UserWarning: Covariance matrix estimation assumes that the model is not penalized. You are estimating a penalized model. The covariance matrix will be incorrect.
  warnings.warn(
python3.9/site-packages/glum/_util.py:37: FutureWarning: is_categorical_dtype is deprecated and will be removed in a future version. Use isinstance(dtype, CategoricalDtype) instead
  if pd.api.types.is_categorical_dtype(dtype) and (column in df)

---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
Cell In[1], line 8
      6 X, y = pd.DataFrame({"x": np.arange(3)}), np.ones(3)
      7 glm = GeneralizedLinearRegressor(family="poisson").fit(X, y)
----> 8 glm.coef_table(X=X, y=y)

File python3.9/site-packages/glum/_glm.py:1397, in GeneralizedLinearRegressorBase.coef_table(self, confidence_level, X, y, mu, offset, sample_weight, dispersion, robust, clusters, expected_information)
   1394     names = self.feature_names_
   1395     beta = self.coef_
-> 1397 covariance_matrix = self.covariance_matrix(
   1398     X=X,
   1399     y=y,
   1400     mu=mu,
   1401     offset=offset,
   1402     sample_weight=sample_weight,
   1403     dispersion=dispersion,
   1404     robust=robust,
   1405     clusters=clusters,
   1406     expected_information=expected_information,
   1407 )
   1409 significance_level = 1 - confidence_level
   1411 std_errors = np.sqrt(np.diag(covariance_matrix))

File python3.9/site-packages/glum/_glm.py:1965, in GeneralizedLinearRegressorBase.covariance_matrix(self, X, y, mu, offset, sample_weight, dispersion, robust, clusters, expected_information, store_covariance_matrix, skip_checks)
   1953 X, y = check_X_y_tabmat_compliant(
   1954     X,
   1955     y,
   (...)
   1961     drop_first=self.drop_first,
   1962 )
[...]

PR #758, opened 2024-01-31
No regularization as default

MatthiasSchmidtblaicherQC:
See discussion on https://github.com/Quantco/glum/issues/728.

Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ Changelog
 3.0.0 - UNRELEASED
 ------------------
 
+**Breaking change:**
+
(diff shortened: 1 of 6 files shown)


lbittarello (approved):

stanmart:
Sorry for the late review, but I've just noticed that [this line](https://github.com/Quantco/glum/blob/137d9fb0f5c815558e2ab53af4591b1a4c65f905/src/glum/_glm.py#L943) became a bit cryptic now that the default alpha is 0. Wouldn't it be a bit clearer if instead of
```python
elif getattr(self, "alpha", 1) == 0 and not self.alpha_search:
```
we did
```python
elif hasattr(self, "alpha") and self.alpha == 0 and not self.alpha_search
```
?

PR #762, opened 2024-02-12
More consistent output from link functions

lbittarello:


lbittarello:
I removed the `__new__` method of `TweedieLink`, because it's quite odd to instantiate a different class for a certain parametrisation and the benefits don't warrant it to my mind. Let me know if I should revert it.

We could also remove the calls to `_asanyarray` and have the link functions preserve the input type, which is consistent in a different way.

lbittarello:
In a similar vein, should we turn `PoissonDistribution`, `GammaDistribution` and `NormalDistribution` into fully fledged classes instead of inheriting from `TweedieDistribution`? The current setup makes it awkward for downstream packages to inherit from them: for example, if I want to modify both `PoissonDistribution` and `TweedieDistribution`,  I either need to live with diamond inheritance (`PoissonDistribution` inherits from `glum.PoissonDistribution`, which inherits from `glum.TweedieDistribution`, and from my `TweedieDistribution`, which also inherits from `glum.TweedieDistribution`) or to following a different pattern from Glum and not have my `PoissonDistribution` inherit from my own `TweedieDistribution`.

jtilly:
> In a similar vein, should we turn PoissonDistribution, GammaDistribution and NormalDistribution into fully fledged classes instead of inheriting from TweedieDistribution?

No objections.

lbittarello:
I'll leave it to a separate PR. :)

jtilly (approved):

PR #765, opened 2024-02-15
Add future warnings about breaking changes in 3.0.0

MatthiasSchmidtblaicherQC:
The warnings reflect the current state of #764 and #677, should be updated accordingly if we make changes.

Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ Unreleased
 - Require Python>=3.9 in line with `NEP 29 <https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table>`_
 - Build and test with Python 3.12 in CI.
 - Added line search stopping criterion for tiny loss improvements based on gradient information.
+- Added warnings about breaking changes in future versions.
 
 2.6.0 - 2023-09-05
 ------------------
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -59,7 +59,7 @@ from ._solvers import (
     _least_squares_solver,
     _trust_constr_solver,
 )
-from ._util import _align_df_categories, _safe_toarray
+from ._util import _align_df_categories, _positional_args_deprecated, _safe_toarray
 
 _float_itemsize_to_dtype = {8: np.float64, 4: np.float32, 2: np.float16}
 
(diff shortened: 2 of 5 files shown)


lbittarello (approved):

stanmart:
Sounds good! What about only showing the warning, if the user tries to invoke the function in a 3.0.0-incompatible way, though? I'm thinking of something like:

```python
    @_positional_args_warning(unchanged_args=("X", "sample_weight", "offset"))
    def predict(
        self,
        X: ShapedArrayLike,
        sample_weight: Optional[ArrayLike] = None,
        offset: Optional[ArrayLike] = None,
        alpha_index: Optional[Union[int, Sequence[int]]] = None,
        alpha: Optional[Union[float, Sequence[float]]] = None,
    ):
    ...
```

where the decorator looks something like this:
```python
def _positional_args_warning(unchanged_args=(), unchanged_args_number=None):
    """Raise a FutureWarning if any positional arguments not in exclude are passed."""
    if unchanged_args_number is None:
        unchanged_args_number = len(unchanged_args)

    def decorator(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            if len(args) > unchanged_args_number + 1:  # +1 for self
                excluded_string = (
                    f"Arguments other than {', '.join(unchanged_args)}"
                    if unchanged_args
                    else "All arguments"
                )
                name = func.__name__ if func.__name__ != "__init__" else args[0].__class__.__name__

                raise FutureWarning(
                    f"{excluded_string} to `{name}` will become keyword-only in 3.0.0."
                )
            return func(*args, **kwargs)

        return wrapper

    return decorator
```

MatthiasSchmidtblaicherQC:
I would be fine with being overabundant and slightly annoying with the `FutureWarning`s, at least they alert the user that there is a major release coming up (and users can always filter them).  Your more targeted approach seems of course superior though and, since you already built it, I think that we should go for it. People who actually want to work with v2.7.0 will thank you. Would you have the time adjust the PR accordingly?

stanmart:
Sure, I can do it tomorrow. It's just a quick proof of concept now but I think this approach should work without many issues.

But I'm also okay with the current more cautious warnings. Maybe that will make more people switch to v3. :) Let me know which one we should go for.

MatthiasSchmidtblaicherQC:
Maybe it is best to just follow what other packages do. `pd.concat` had a similar breaking change in 2.0.0, which they solved by... [a decorator called `deprecate_nonkeyword_arguments`](https://github.com/pandas-dev/pandas/blob/573e7eaffd801ee5bd1f7685697b51eef5b8ed85/pandas/core/reshape/concat.py#L147) which looks quite similar to what you are suggesting :) and raises `FutureWarning: In a future version of pandas all arguments of concat except for the argument 'objs' will be keyword-only`. 

So I guess we should go with the decorator approach. We could just import the one from pandas, but, if I see correctly, it does take into account reshuffling of arguments as in the `covariance_matrix` case.

stanmart:
Oops, I might have reinvented the wheel... I made the changes. I think it works, but could you please also check if it does what it's supposed to do?

MatthiasSchmidtblaicherQC:
Thanks a lot! I tested and it works as expected for me.

PR #768, opened 2024-02-21
Restructure distributions

lbittarello:
This PR restructures the distributions to avoid diamond inheritance -- i.e. the gamma, inverse Gaussian, normal and Poisson distributions no longer inherit from the Tweedie distribution. Because it is often convenient to test whether a distribution belongs to the Tweedie family, I added a `to_tweedie` method to the base class, which allows users to recover the Tweedie representation of a family if it exists. (I added an equivalent method to `Link`.)

Because it is a breaking change, the base branch is `glum_v3`.

Some free riders:

- `ExponentialDispersionModel.in_y_range` has been cleaned up.
- The documentation of the different distributions has been cleaned up and simplified.
- Excessively restrictive type hints have been removed. A few new hints have been added.
- Cython functions for the inverse Gaussian distribution have been added.
- The parsing of Tweedie links in `get_family` has been aligned on `get_family`.

### DIFF
--- a/src/glum/_distribution.py
+++ b/src/glum/_distribution.py
@@ -14,6 +14,10 @@ from ._functions import (
     gamma_log_eta_mu_deviance,
     gamma_log_likelihood,
     gamma_log_rowwise_gradient_hessian,
+    inv_gaussian_deviance,
+    inv_gaussian_log_eta_mu_deviance,
+    inv_gaussian_log_likelihood,
+    inv_gaussian_log_rowwise_gradient_hessian,
     negative_binomial_deviance,
     negative_binomial_log_eta_mu_deviance,
     negative_binomial_log_likelihood,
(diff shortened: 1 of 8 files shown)


src/glum/_distribution.py
```diff
@@ -306,46 +273,41 @@ def deviance_derivative(self, y, mu, sample_weight=1):
...
-        Compute:
-        * the linear predictor, ``eta``, as ``cur_eta + factor * X_dot_d``;
-        * the link-function-transformed prediction, ``mu``;
-        * the deviance.
+        cur_eta,
+        X_dot_d,
+        y,
+        sample_weight,
+    ) -> tuple[np.ndarray, np.ndarray, float]:
+        """Compute ``eta``, ``mu`` and the deviance.
```
MatthiasSchmidtblaicherQC, line 305:
```suggestion
        """Compute ``eta``, ``mu``, and the deviance.
```

src/glum/_distribution.py
```diff
@@ -1067,19 +1084,251 @@ def dispersion(self, y, mu, sample_weight=None, ddof=1, method="pearson") -> flo
...
+    :class:`~glum.ExponentialDispersionModel`, for details.
+    """
+
+    lower_bound = -np.inf
+    upper_bound = np.inf
+    include_lower_bound = False
+    include_upper_bound = False
+
+    def __eq__(self, other):  # noqa D
+        return isinstance(other, self.__class__)
```
MatthiasSchmidtblaicherQC, line 1239: `self.__eq__` is identical for all distributions except for Tweedie and negative binomial distributions. Should we move the method to the superclass?

MatthiasSchmidtblaicherQC (approved):
Thanks for the cleanup. I have some minor suggestions.

src/glum/_distribution.py
```diff
@@ -104,51 +86,39 @@ def include_upper_bound(self) -> bool:
...
+        else:
+            ub_op = np.less
+
+        return lb_op(x, self.lower_bound) & ub_op(x, self.upper_bound)
+
+    def to_tweedie(self):
+        """Return the Tweedie representation of a distribution if it exists."""
+        if hasattr(self, "__tweedie_repr__"):
+            return self.__tweedie_repr__()
+        return None
```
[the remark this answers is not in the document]
lbittarello: I added this method to facilitate checks [like this](https://github.com/Quantco/glum/pull/768/files#diff-ec36b874d09fcf9c0144f5796ac1fa9aa16a54bb0017b65fbc75d64a12c4269bR507). If we raise, we'll need to catch the error, which is slow and ungainly. Would it help if we added an argument (`safe`?) to control whether to raise?

src/glum/_distribution.py
lbittarello: The Oxford comma is optional. ;) And superfluous too.

src/glum/_distribution.py
lbittarello: We could. My concern is that someone could implement a child class and end up with a nonsensical equality check by accident.

src/glum/_distribution.py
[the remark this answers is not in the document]
MatthiasSchmidtblaicherQC: I see. I think that raising by default if there is no Tweedie representation, but disabling the error in the checks via a `safe` argument sounds like a good solution.

src/glum/_glm.py
```diff
@@ -3669,42 +3669,29 @@ def _compute_information_criteria(
...
-        if (
-            isinstance(self.family_instance, TweedieDistribution)
-            and self.family_instance.power > 2
-        ):
-            raise NotImplementedError(model_err_str)
+        if not hasattr(self.family_instance, "log_likelihood"):
+            raise NotImplementedError(
+                "The family instance does not define a `log_likelihood` method, so "
+                "information criteria cannot be computed. Compatible families include "
+                "the binomial, negative binomial and Tweedie (power<2 or power=3)."
```
MatthiasSchmidtblaicherQC, line 3676: Nitpicking, but shouldn't it be
```suggestion
                "the binomial, negative binomial and Tweedie (power<=2 or power=3)."
```
?

Issue #775, opened 2024-03-04
Clarify effect of `scale_predictors` on penalization in docstring


mayer79:
When using penalized regression, by default, H2o applies penalties to internally standardized features. This is convenient because users do not need to scale the input manually to ensure fair application of penalties. 

How is it done with glum? It seems that features are internally scaled. Are penalties applied to standardized coefficients or to the original coefficients?

MatthiasSchmidtblaicherQC:
Whether penalties are applied to standardized features is governed by the `scale_predictors` argument, see [here](https://github.com/Quantco/glum/blob/d65ac0d7305e408224c062d57f80389c3fb638f8/src/glum/_glm.py#L2984). To be precise, features are always standardized, but, in case that `scale_predictors=False`, the penalty is adjusted as if there was no standardization.

mayer79:
Oh, I see, thanks!

I am typically interested in getting unscaled coefficients in ` coef_table()`, and at the same time applying penalties to an (internally) scaled design matrix to ensure that all features are treated equally. If I got the docu of `scale_predictors` right, this combination is not easily available?

MatthiasSchmidtblaicherQC:
No problem! 

The behavior that you describe should be the default, the coefficients are ["unstandardized" again](https://github.com/Quantco/glum/blob/d65ac0d7305e408224c062d57f80389c3fb638f8/src/glum/_glm.py#L3049). You can also check this with an example:
```
import pandas as pd
import numpy as np

from glum import GeneralizedLinearRegressor

rng = np.random.default_rng(42)

X = pd.DataFrame({"x1": rng.normal(size=100), "x2": rng.normal(size=100)})
y = df["x1"] + df["x2"] + rng.normal(scale=0.1, size=100)
X["x2"] *= 100 

GeneralizedLinearRegressor(alpha=0.1, scale_predictors=True).fit(X, y).coef_ # second coef is smaller by order of -2
```

> what does "adjusted as if..." mean?

The standardization is also applied to the penalty matrix so that score equations are unchanged up to an affine transformation compared to no standardization (this is done for numerical stability). You can have a look at `_standardize` for the details if you are interested.

mayer79:
Fantastic, thanks a lot for the explanation and the example!

mayer79:
Should we mention in the docu of `scale_predictors` that the default (`False`) will scale the penalties, i.e., if one wants to penalize each feature equally, one would need to set `scale_predictors=True`?

[...]

PR #778, opened 2024-03-07
Explain `scale_predictors` more

MatthiasSchmidtblaicherQC:
See https://github.com/Quantco/glum/issues/775.

### DIFF
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -3127,10 +3127,12 @@ class GeneralizedLinearRegressor(GeneralizedLinearRegressorBase):
         set ``verbose`` to any positive number for verbosity.
 
     scale_predictors: bool, optional (default=False)
-        If ``True``, estimate a scaled model where all predictors have a
-        standard deviation of 1. This can result in better estimates if
-        predictors are on very different scales (for example, centimeters and
-        kilometers).
+        If ``True``, scale all predictors to have standard standard deviation one.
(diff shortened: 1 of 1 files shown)


src/glum/_glm.py
```diff
@@ -3127,10 +3127,12 @@ class GeneralizedLinearRegressor(GeneralizedLinearRegressorBase):
         set ``verbose`` to any positive number for verbosity.
 
     scale_predictors: bool, optional (default=False)
-        If ``True``, estimate a scaled model where all predictors have a
-        standard deviation of 1. This can result in better estimates if
-        predictors are on very different scales (for example, centimeters and
-        kilometers).
+        If ``True``, scale all predictors to have standard standard deviation one.
```
jtilly, line 3130:
```suggestion
        If ``True``, scale all predictors to have standard deviation one.
```

jtilly (approved):

src/glum/_glm.py
```diff
@@ -3127,10 +3127,12 @@ class GeneralizedLinearRegressor(GeneralizedLinearRegressorBase):
...
+        This can result in better estimates if predictors are on different scales
+        (for example, centimeters and kilometers). Should be set to ``True`` if
```
mayer79, line 3132: Can we even remove this? "This can result in better estimates if predictors are on different scales (for example, centimeters and kilometers)." (Since, internally, the design matrix is scaled anyway?)

PR #780, opened 2024-03-11
Accept `None` more generally

lbittarello:
For some reason, Glum requires many arguments to be strings where `None` would have been a more obvious default. This PR tweaks the code to accept `None` as well, though it doesn't change the default values.

It also removes the leading underscores from `_dtype` and `_stype`. Not sure why they were ever there.

### DIFF
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -2144,14 +2144,14 @@ class GeneralizedLinearRegressorBase(BaseEstimator, RegressorMixin):
                 https://github.com/scikit-learn/scikit-learn/pull/9405.
                 """
             )
-        if self.solver not in {
+        if self.solver not in [
             None,
             "auto",
             "irls-ls",
             "lbfgs",
             "irls-cd",
             "trust-constr",
-        }:
+        ]:
             raise ValueError(
                 "GeneralizedLinearRegressor supports only solvers"
                 " 'auto', 'irls-ls', 'lbfgs', 'irls-cd' and 'trust-constr'; "
@@ -2182,7 +2182,7 @@ class GeneralizedLinearRegressorBase(BaseEstimator, RegressorMixin):
             raise TypeError(
                 f"The argument warm_start must be bool; got {self.warm_start}."
             )
-        if self.selection not in {"cyclic", "random"}:
+        if self.selection not in ["cyclic", "random"]:
             raise ValueError(
                 "The argument selection must be 'cyclic' or 'random'; "
                 f"got {self.selection}."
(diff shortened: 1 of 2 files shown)


MarcAntoineSchmidtQC:
Could you elaborate on why you think it's more obvious to accept None? I'm happy to accept this PR (after we add the proper documentation) but I'm curious to see from a user point-of-view why this makes more sense.

The obvious downside to this PR is that it makes the code more verbose and bloated, and requires more documentation to explain what None does.

src/glum/_glm.py
```diff
@@ -533,37 +538,41 @@ def setup_p1(
 
     # P1 and P2 are now for sure copies
     P1 = alpha * l1_ratio * P1
-    return cast(np.ndarray, P1).astype(_dtype)
+    return cast(np.ndarray, P1).astype(dtype)
 
 
 def setup_p2(
-    P2: Union[str, np.ndarray, sparse.spmatrix],
+    P2: Optional[Union[str, np.ndarray, sparse.spmatrix]],
```
MarcAntoineSchmidtQC, line 545: In a month or so, when we drop support for python 3.9 we will be able to use the much cleaner (IMO) `type | None`.

src/glum/_glm.py
```diff
@@ -2137,7 +2144,14 @@ def _validate_hyperparameters(self) -> None:
                 https://github.com/scikit-learn/scikit-learn/pull/9405.
                 """
             )
-        if self.solver not in ["auto", "irls-ls", "lbfgs", "irls-cd", "trust-constr"]:
+        if self.solver not in [
+            None,
```
MarcAntoineSchmidtQC, line 2148: We should add this behavior to the documentation.

MarcAntoineSchmidtQC (changes requested):
My only request before fully accepting is that we change the documentation to explain what None does.

lbittarello:
> Could you elaborate on why you think it's more obvious to accept ``None``?

My feeling is that `None` is generally understood to mean "unspecified", hence "default".

Scikit-learn seems to be inconsistent in its choice of default values. For example, the [`AgglomerativeClustering`](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html#sklearn.cluster.AgglomerativeClustering) uses `"auto"` as the default for `compute_full_tree` and `None` as the default for `memory`, `connectivity` and `distance_threshold`.

Upon reflection, `None` is ambiguous as a value for `link` (does it mean no link, i.e. the identity link?) and `solver` (no `solver`? closed-form solutions only?), so I reverted those changes and only kept `None` as an option for the penalty matrices, where it should be obvious.

PR #782, opened 2024-03-20
Move helpers into `_utils`

lbittarello:
A little clean-up:

- `_expand_categorical_penalties` doesn't leverage the scope at all, so it doesn't need to be an inner function.
- `_is_contiguous` is self-standing, so it doesn't need to be a method.
- No reason to use `chain` anywhere.

### DIFF
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -20,7 +20,6 @@ import re
 import sys
 import warnings
 from collections.abc import Iterable, Mapping, Sequence
-from itertools import chain
 from typing import Any, NamedTuple, Optional, Union, cast
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


MatthiasSchmidtblaicherQC (approved):
Thanks for spotting and taking care of this!

src/glum/_util.py
```diff
@@ -114,6 +114,54 @@ def _add_missing_categories(
     return df
 
 
+def _expand_categorical_penalties(
+    penalty, X, drop_first, has_missing_category
+) -> np.ndarray:
+
+    # If `P1` or `P2` has the same shape as `X` before expanding
+    # categorical columns, we assume that the penalty at the location of
+    # categorical columns is the same for all levels.
```
lbittarello, line 123:
```suggestion
```

src/glum/_glm.py
```diff
@@ -2679,6 +2675,7 @@ def _set_up_and_check_fit_args(
         P2 = self.P2
 
         copy_X = self._should_copy_X()
+        drop_first = getattr(self, "drop_first", False)
```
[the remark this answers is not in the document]
lbittarello: It is. :)

src/glum/_util.py
```diff
@@ -114,6 +114,54 @@ def _add_missing_categories(
     return df
 
 
+def _expand_categorical_penalties(
+    penalty, X, drop_first, has_missing_category
+) -> np.ndarray:
+"""Determine penalty matrices`P1` or `P2` after expanding categorical columns.
```
MatthiasSchmidtblaicherQC, line 120:
```suggestion
"""Determine penalty matrices `P1` or `P2` after expanding categorical columns.
```

PR #792, opened 2024-05-21
Fix `test_formula` for pandas 3.0 nightly

stanmart:
Checklist

This change should fix the `formulaic.errors.FormulaMaterializerNotFoundError: No materializer has been registered for input type 'pandas.DataFrame'. Available input types are: set().` errors in the nightlies run. The underlying cause is that the current version of formulaic has some issues with pandas 3.0, and `model_matrix` does not work without selecting a materializer explicitly.

IMO being explicit is not a bad idea anyway, so we can fix the test instead of waiting for a formulaic update.

### DIFF
--- a/tests/glm/test_glm.py
+++ b/tests/glm/test_glm.py
@@ -1716,7 +1716,9 @@ def test_column_with_stddev_zero():
 
     model = GeneralizedLinearRegressor(
         family="poisson", fit_intercept=False, scale_predictors=False
-    ).fit(X, y)  # noqa: F841
+    ).fit(
+        X, y
+    )  # noqa: F841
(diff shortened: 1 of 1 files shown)


MatthiasSchmidtblaicherQC:
Thanks! Just FYI, the next formulaic release should fix the underlying issue [as well](https://github.com/matthewwardrop/formulaic/issues/185).

MatthiasSchmidtblaicherQC:
Second FYI: nightlies should still fail because of [tabmat](https://github.com/Quantco/tabmat/issues/357).

MatthiasSchmidtblaicherQC (approved):

PR #795, opened 2024-05-23
Fix sklearn C type change

MarcAntoineSchmidtQC:
We were using an internal function from sklearn. `_cd_fast` is already a module copied and modified from sklearn. I did the same with the RNG function.

Checklist

### DIFF
--- a/src/glum/_cd_fast.pyx
+++ b/src/glum/_cd_fast.pyx
@@ -17,21 +17,39 @@ from cython.parallel import prange
 import warnings
 from sklearn.exceptions import ConvergenceWarning
 
-from sklearn.utils._random cimport our_rand_r
-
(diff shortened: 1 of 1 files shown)


src/glum/_cd_fast.pyx
```diff
@@ -17,23 +17,41 @@ from cython.parallel import prange
...
 
 np.import_array()
 
-
-# The following two functions are shamelessly copied from the tree code.
-
+# The following two functions are shamelessly copied from the tree code. (_random.pxd)
+# Authors: Arnaud Joly
+#
+# License: BSD 3 clause
```
xhochy, line 28:
```suggestion
# License: BSD-3-clause
```

jtilly (approved):

Issue #801, opened 2024-05-29
Use pixi for setting up the dev environment


stanmart:
At the moment, the [contribution guidelines](https://glum.readthedocs.io/en/latest/contributing.html) suggest setting up a conda environment based on the `environment.yml` files in the repo. I think there is quite a lot to gain from moving to pixi in the case of glum. In particular:

 - There are various `environment.yml` files for different use cases. Pixi has an elegant way of handling [multiple environments](https://pixi.sh/latest/features/multi_environment/)
 - Certain dependencies are platform specicfic, which pixi also [handles nicely](https://pixi.sh/latest/features/multi_platform_configuration/) (see #787)
 - It uses [lockfiles](https://pixi.sh/latest/features/lockfile/), which would help ensure that everyone (including the CI) uses the same versions of the dependencies. Especially useful for benchmarking.
 - I personally find it more ergonomic than activating environments but YMMV

What would not change:

 - Pixi still uses conda packages for dependencies
 - The build system. Pixi can not build conda packages (yet?). CI conda builds would still be based on `conda.recipe/meta.yaml` and  `conda-mambabuild`, and PyPI builds on `cibuildwheel`.

I can make a draft PR so we get a better idea how it would look like. What do you think?

jtilly:
I'm in favor of moving things to pixi. This is a nice repo to try things out. 

It would be great if we could address the following items:
- we need to regularly (and automatically) update the lockfiles so that we're testing against the most recent versions of our dependencies
- we should also add a test against "oldies", i.e. the minimal dependencies that we want to support
- it would be nice if we could get the tests against [nightlies](https://github.com/Quantco/glum/blob/15dda3b72ffc3c31bf151244f67046deb2e48305/.github/workflows/daily.yml#L37-L51) also included in the pixi setup

stanmart:
Good points. Let's continue under #802

PR #802, opened 2024-05-30
Switch to pixi

stanmart:
This is a proposal to move from the current conda/mamba based dev installation to pixi. See #801 Check out the updated contributing page of the docs for info on how it changes the setup process.

Checklist

TODO:
* [x] Make CI work with pixi
  * [x] Use the [setup-pixi](https://github.com/marketplace/actions/setup-pixi) gh action
  * [x] Add features with different python versions
  * [x] Doctests
  * [x] Pre-commit
* [x] Make daily runs work with pixi
  * [x] See if tests against nightlies can be incorporated into an environment in pixi.toml
* [x] Platform dependent deps (e.g. intel-mkl for win)
* [x] Add other platforms to pixi.toml
* [x] Commit lockfile when we are happy with the basics
* [x] Figure out how to automatically update lockfile periodically (could be e.g. some weekly CI action)
* [x] Add environment with minimal dependencies that we want to support, and test against these "oldies"
* [x] Make it work with readthedocs

Advantages:
 - Can guarantee that local dev environment is the same for everyone (on the same platform), and also in CI
 - Can use platform-dependent requirements (e.g. intel-mkl for the benchmark env on non-arm archs)
 - Super fast
 - One file containing multiple environments, instead of having multiple `environment*.yml` files and having to do `conda env update`s to create certain environments.
 - Includes environments and tasks to test on oldest supported deps (oldies) and nightlies, both locally and in CI
 - There are pre-defined tasks for lots of commonly-used tasks (e.g. make and serve docs, install glum in editable mode, run tests).

Limitations:
 - Whille pixi supports editable installs, it cannot pass the necessary flags to pip/uv. This is the reason why we need the annoying postinstall task. (Not any worse than the current status quo, though.) Some proposed pixi features might help with this in the future:
   - [lifecycle hooks](https://github.com/prefix-dev/pixi/issues/1183)
   - [exposing more pip/uv flags](https://github.com/prefix-dev/pixi/issues/1124)
 - pixi is still pre-1.0
 - Most people have some flavor of conda installed, but pixi is less widespread


stanmart:
@pavelzw, let me know if you have any input regarding this PR or the general approach

pixi.toml
```diff
@@ -0,0 +1,92 @@
...
+shapely = "*" # used in docs/tutorials/regularization_housing_data
+libpysal = "*" # used in docs/tutorials/regularization_housing_data
+
+[feature.benchmark.dependencies]
+h2o-py = "*"
+openjdk = "*"
+
+[environments]
+default = ["dev", "test"]
+docs = ["dev", "docs", "tutorial"]
```
pavelzw, line 91: Why is dev needed in docs? I would suggest only using docs to build the docs and nothing else.

pixi.toml
```diff
@@ -0,0 +1,92 @@
+[project]
+name = "glum"
+channels = ["conda-forge"]
+platforms = ["linux-64"]
```
pavelzw, line 4: Let's add osx-arm64 and win-64 as well (and run tests in ci)

pixi.toml
```diff
@@ -0,0 +1,92 @@
...
+[feature.benchmark.tasks]
+postinstall-benchmark = "pip install --no-use-pep517 --no-build-isolation --no-deps --disable-pip-version-check -e ."
+glm-benchmarks-run = "glm_benchmarks_run"
+glm-benchmarks-analyze = "glm_benchmarks_analyze"
+
+[build-dependencies]
+c-compiler = "*"
+cxx-compiler = "*"
+cython = "*"
+setuptools_scm = ">=8.1"
```
pavelzw, line 32:
```suggestion
setuptools-scm = ">=8.1"
```

pavelzw:
> Figure out how to automatically update lockfile periodically (could be e.g. some weekly CI action)

We already have this feature in our organization since last week.
An automatic PR comes every month and runs `pixi update`.

You can specify its behavior by setting the following in pixi.toml

```

[tool.update]
autoupdate-branch = "pixi-update"
autoupdate-commit-message = "Update pixi lockfile"
autoupdate-pull-request-labels = ["dependencies"]
autoupdate-schedule = "monthly" # weekly, monthly, quarterly, never
ignore-environments = []
ignore-platforms = []
```

pavelzw:
> See if tests against nightlies can be incorporated into an environment in pixi.toml

Would be interesting to see if/how this is supported the best way with pixi. Maybe running `pixi remove dependency && pixi add --pypi --nightly(?) dependency` in CI?

pavelzw (commented):
Efforts look good so far 👍

.github/workflows/ci.yml
```diff
@@ -2,76 +2,50 @@ name: CI
...
+  #     - name: Install Python 3.11
+  #       uses: actions/setup-python@v5
+  #       with:
+  #         python-version: "3.11"
+  #     - name: Run pre-commit checks
+  #       uses: pre-commit/action@v3.0.1
 
   unit-tests:
-    name: Tests - ${{ matrix.os }} - Py${{ matrix.python-version }}
+    name: Tests - ${{ matrix.os }} - ${{ matrix.environment }}
```
pavelzw, line 26:
```suggestion
    name: Tests
```

This would render the name to be 

`Tests (ubuntu-latest, py39)`

pixi.toml
```diff
@@ -0,0 +1,123 @@
...
+py312 = ["py312", "test"]
+oldies = ["oldies", "test"]
+
+[tool.update]
+autoupdate-branch = "pixi-update"
+autoupdate-commit-message = "Update pixi lockfile"
+autoupdate-pull-request-labels = ["dependencies"]
+autoupdate-schedule = "monthly" # weekly, monthly, quarterly, never
+ignore-environments = []
+ignore-platforms = []
```
pavelzw, line 123: These are the default values. If you don't plan on specifying them, I would just suggest leaving them out. The updater will come even if you don't specify `[tool.update]`.

pixi.toml
stanmart: The idea was that developing the documentation can involve working on the tutorial notebooks, and those tools can be helpful there. They are probably not, though :)

stanmart:
Thanks so much for the help @pavelzw! I think it is shaping up nicely. It has already reveled that our minimum pinned versions are not recent enough.

When we figure out how to handle nightlies (should be doable) and read the docs (less sure about this one) we should be good.

pavelzw:
There is a more-or-less hacky way to do readthedocs with pixi, see https://github.com/prefix-dev/pixi/pull/1423/files#diff-65d574b20be487abb02f3d039096768be5613f955410494c7862d3f6b3781432 the readthedocs-override example

stanmart:
> > See if tests against nightlies can be incorporated into an environment in pixi.toml
> 
> Would be interesting to see if/how this is supported the best way with pixi. Maybe running `pixi remove dependency && pixi add --pypi --nightly(?) dependency` in CI?

Atm I just use pip install to replace some packages with the nightly versions, without caring about their dependencies. This mirrors the current state of the daily ci runs. But yeah, would be interesting to see if we can do better in the future using pixi.

stanmart:
> There is a more-or-less hacky way to do readthedocs with pixi, see https://github.com/prefix-dev/pixi/pull/1423/files#diff-65d574b20be487abb02f3d039096768be5613f955410494c7862d3f6b3781432 the readthedocs-override example

Yeah it's indeed a bit hacky, but should be fine. I'm wondering if exporting the docs environment into a conda environment.yml, and using that for rtd would be a viable alternative. (I know that there is no pixi export command as of yet, but there [seems to be one](https://github.com/prefix-dev/pixi/pull/1427) on the horizon, and the current pr is not urgent at all.)

pavelzw:
Performance wise installing pixi with conda and then running pixi install is probably still faster 🤔

.github/workflows/daily.yml
```diff
@@ -9,62 +9,52 @@ on:
...
-          - PYTHON_VERSION: '3.10'
-            NOTE: 'Default Builds' # run once with normal dependencies
+          - environment: 'nightly'
+            NOTE: 'Python 3.12' # run once with normal dependencies
     steps:
       - name: Checkout branch
         uses: actions/checkout@v4
-      - uses: mamba-org/setup-micromamba@f8b8a1e23a26f60a44c853292711bacfd3eac822
+      - name: Set up pixi
+        uses: prefix-dev/setup-pixi@v0.8.0
```
pavelzw, line 28: let's keep this pinned to the SHA, dependabot can still update it

pixi.toml
```diff
@@ -0,0 +1,134 @@
...
+c-compiler = "*"
+cxx-compiler = "*"
+cython = "!=3.0.4"
+make = "*"
+mako = "*"
+xsimd = "<11|>12.1"
+[feature.build-tabmat.target.linux-64.dependencies]
+jemalloc-local = "*"
+[feature.build-tabmat.target.osx-arm64.dependencies]
+jemalloc-local = "*"
```
pavelzw, line 103:
```suggestion
[feature.build-tabmat.target.unix.dependencies]
jemalloc-local = "*"
```

stanmart:
Should we also add a nighlies test to the ci that runs on pushes?

pavelzw (approved):

MatthiasSchmidtblaicherQC:
Sounds promising! I have yet to work with Pixi and cannot comment on the details. Regarding:

> Should we also add a nighlies test to the ci that runs on pushes?

Can we run nightlies for pushes only in PRs outside of draft mode? That would allow one to push small incremental commits without overwhelming the CI during development while being able to notice nightly failures earlier.

stanmart:
Yeah that's a good idea. Right now I think the CI runs on every push, not even only on PRs. We could have a set of leaner checks (e.g. one linux environment plus pre-commit), and all the rest could be run only on non-draft PRs. One limitation is that [it seems that](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) workflows dispatched by pull request related events do not trigger if there are merge conflicts.

pavelzw:
In general it is better for public repos to use something like

```yaml
on:
  push:
    branches: [main]
  pull_request:
```

as otherwise the CI will only run when pushed to a branch in `quantco/glum`, not when pushed to a branch from a fork.

MarcAntoineSchmidtQC:
quick question for pixi expert: is there an equivalent to doing `conda env update`? My use case is that I usually have glum AND tabmat installed in editable mode in the same environment. I use `conda env update` to install both development dependencies in the same environment. Is that possible with pixi?

MarcAntoineSchmidtQC:
Regarding the nightly on all pushes: One downside I've seen in the past is that nightlies are by design more fragile and can fail for multiple reasons outside of our control. We have seen the nightly fail for long periods of time until someone has the time to fix the future deprecation. I would not want to slow down the development of unrelated features because the nightly has been failing for a month.

However, I am not against the idea, but want to contribute to the discussion with past experiences.

pavelzw:
> is there an equivalent to doing conda env update

there is `pixi update` which updates your lockfile as well as your environment.

> I usually have glum AND tabmat installed in editable mode in the same environment

This is not directly possible with this PR: here, `tabmat` is a direct dependency and always being installed in non-editable mode.

You could install tabmat in editable mode in this way:

```toml
[dependencies]
numexpr = "*"
# no tabmat here

[pypi-dependencies]
tabmat = { path = "../tabmat", editable = true }
```

@stanmart maybe it might make sense to have a separate `dev` environment where we have tabmat as a pypi editable dependency (or not at all and instead use something like `postinstall-tabmat`) and in the default envs, we get it from `conda-forge`.

> I use conda env update to install both development dependencies in the same environment

not sure i follow 🤔 but that's probably because i'm not familiar with the in's and out's of `conda env update`...

---

In general, what you are describing sounds somewhat similar to a workspace feature that is being discussed but not yet realized:

- https://github.com/prefix-dev/pixi/issues/1285#issuecomment-2084734687
- https://github.com/prefix-dev/pixi/discussions/387

stanmart:
There is already a feature for [tabmat's build dependencies](https://github.com/Quantco/glum/pull/802/files#diff-b092682f0ad15a352f9fa2b0a67c992454b586275ee76b74a5839ab9afdf9894R108-R116), so we could add a new environment, such as

```
glum-tabmat = ["dev", "build-tabmat", "test"]
```

(we could also include the benchmarks deps in this env if that's useful), plus a new task for the build-tabmat feature:

```
[feature.build-tabmat.tasks.install-editable]
"""
pip install --no-use-pep517 --no-deps -e
"""
```

Then one could set up a combined glum-tabmat environment like so:
```bash
pixi run -e glum-tabmat install-editable /path/to/tabmat && pixi run -e glum-tabmat postinstall
```

Then tests could be run by doing
```bash
pixi run -e glum-tabmat test
```

or a shell with an activated environment  can be launched by
```bash
pixi shell -e glum-tabmat
```

Would that work for your use case @MarcAntoineSchmidtQC?

stanmart:
> ```toml
> [pypi-dependencies]
> tabmat = { path = "../tabmat", editable = true }
> ```

The issue is that installing tabmat requires the `--no-use-pep517` and  `--no-deps` flags, so AFAIK we cannot (yet) use `[pypi-dependencies]` to add it. I agree though that it would be ideal.

> (or not at all and instead use something like `postinstall-tabmat`)

Yeah I think this is a good solution. A question regarding this: is there any harm in doing it in the default env where tabmat is already installed via `pixi`? I believe `pip` simply uninstalls it before installing the editable version.

pavelzw:
> is there any harm in doing it in the default env where tabmat is already installed via pixi? I believe pip simply uninstalls it before installing the editable version.

Whenever pixi.lock changes, pixi will probably re-install tabmat from conda. So i think the cleanest way as of now would be to have a separate environment where tabmat is just not installed.

stanmart:
> Whenever pixi.lock changes, pixi will probably re-install tabmat from conda. So i think the cleanest way as of now would be to have a separate environment where tabmat is just not installed and then run `postinstall-tabmat` there.

Good point. The latest commit addresses this requirement then. You can simply run
```bash
pixi run postinstall-glum-tabmat
```
after which you will have an environment called `glum-tabmat` with editable installs for both packages. The command assumes that tabmat's folder is at `../tabmat` relative to glum.

Btw I cannot wait for postinstall hooks and/or better editable installs, because even though it is probably the best solution, it still feels ugly with the almost-duplicated `default` and `no-tabmat` features.

MarcAntoineSchmidtQC:
Thanks! This perfectly answers my question. Perfectly happy to merge this PR when ready.

MarcAntoineSchmidtQC (approved):

pixi.toml
```diff
@@ -0,0 +1,163 @@
...
+pytest = "*"
+pytest-xdist = "*"
+git_root = "*"
+click = "*"
+attrs = "*"
+psutil = "*"
+mypy = "*"
+
+[feature.dev.dependencies]
+ruff = "*"
```
jtilly, line 83: Why do we need ruff here?

pixi.toml
```diff
@@ -0,0 +1,163 @@
...
+ipdb = "*"
+ipython = "*"
+line_profiler = "*"
+memory_profiler = "*"
```
jtilly, line 87:
```suggestion
memory_profiler = "*"
ipykernel = "*"
```
Can we add ipykernel here for vscode users (or should that go somewhere else)?

pixi.toml
stanmart: Good point, we don't

stanmart:
There is one final question on my part about the mypy pre-commit hook. In the proposed version, we are running it in an [environment](https://github.com/Quantco/glum/pull/802/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R33) where glum and all of its deps are installed. As a result, it finds lots of errors (most of which are not problematic ofc) even if only a tiny part of the source code changes. While this would be fine for a new project, it would require us to sprinkle `type: ignore`s all over the code, and make sure mypy has no problems with even the unmodified parts of the codebase.

An alternative would be to move mypy into the lint environment, which would be analogous to the way things worked before (pre-commit's mypy in a separate virtual env, no access to imports). It would come up with fewer false positives at the cost of being able to catch fewer errors.

What should we do about this?

stanmart:
@MatthiasSchmidtblaicherQC I like your idea about running just a subset of checks while a PR is draft (or the branch does not have an associated pr), but I feel that that is somewhat orthogonal to the current PR. I'd rather limit this PR to changing the package manager while keeping other things constant, and create a separate issue to improve the ci pipeline. Would that be okay with you?

MarcAntoineSchmidtQC:
@stanmart, I'm a bit confused. Maybe something that would help me right now is if you can point me to an example of a failed mypy run due to this new setup? Can I find it in the CI?

I ran `pixi run pre-commit-run` in the new environment, and I get a bunch of errors, but they are all internal error to glum (not from the dependencies). Are those the errors you are talking about?

.pre-commit-config.yaml
```diff
@@ -1,24 +1,36 @@
...
-          - --namespace-packages
-        exclude: (^tests/|^src/glum_benchmarks/orig_sklearn_fork/)
-        additional_dependencies: [-c, conda-forge, types-setuptools=67.5, attrs]
-  - repo: https://github.com/Quantco/pre-commit-mirrors-cython-lint
-    rev: 0.16.2
-    hooks:
-      - id: cython-lint-conda
-        args: [--no-pycodestyle]
-      - id: double-quote-cython-strings-conda
+      # ruff
```
pavelzw, line 4:
```suggestion
      # ensure pixi environments are up to date
      # workaround for https://github.com/prefix-dev/pixi/issues/1482
      - id: pixi-install
        name: pixi-install
        entry: pixi install -e default -e lint
        language: system
        always_run: true
        require_serial: true
        pass_filenames: false
      # ruff
```

please also add this for now until https://github.com/prefix-dev/pixi/issues/1482 is fixed

otherwise you will run into strange errors in pre-commit

.pre-commit-config.yaml
```diff
@@ -1,24 +1,36 @@
 repos:
```
pavelzw, line 1:
```suggestion
exclude: ^.pixi$
repos:
```

from my experience, `pre-commit run --all` tries to lint _all_ files and ignores the `.gitignore`. could be wrong, though

pavelzw:
> Not sure why they were not caught before

before, mypy just skipped all `import` statements where the packages were not installed (because of https://github.com/Quantco/glum/blob/65073148c4959060e0f2fbe191b801d3ad2fceae/.pre-commit-config.yaml#L15). in the pre-commit setup from before, mypy was installed in an isolated environment with nothing else in it.

now, mypy has these packages available (since it's installed in the `default` environment) and can actually check whether the types are correct or not

MarcAntoineSchmidtQC:
@stanmart, I added two commits. One to fix the "exclude" option of mypy which needs to be specified in the pre-commit-hook file and not in the config (a quirk of pre-commit). The other to fix the mypy errors. I get all green on my end when running locally.

MarcAntoineSchmidtQC:
Final point: I think with this new setup we may need to run `pre-commit run --all`, and not simply `pre-commit run`. My understanding of the github action (which we previously used) is that it was running on all files. When running as a pre-commit hook, it only runs on changes, but the CI runs on everything.

Fixing this will run the pre-commit checks in the CI instead of skipping them. See [here for example](https://github.com/Quantco/glum/actions/runs/9488749701/job/26148408317?pr=802#step:4:11).

pavelzw:
> One to fix the "exclude" option of mypy which needs to be specified in the pre-commit-hook file and not in the config (a quirk of pre-commit).

It's true that you need to add it to `.pre-commit-config.yaml` but you might want to reconsider removing the exclude from pyproject.toml (I know, it's bad to repeat oneself) because this will lead to `pixi run mypy .` behaving differently than running mypy within the pre-commit hook.

.pre-commit-config.yaml
```diff
@@ -1,24 +1,37 @@
...
+      - id: cython-double-quote
+        name: cython-double-quote
+        entry: pixi run -e lint double-quote-cython-strings
+        language: system
+        types: [cython]
+        require_serial: true
+      # mypy
+      - id: mypy
+        name: mypy
+        entry: pixi run -e default mypy --check-untyped-defs --ignore-missing-imports --namespace-packages
```
pavelzw, line 33:
```suggestion
        entry: pixi run -e default mypy
```

I would suggest moving `--check-untyped-defs` and `--namespace-packages` to the `pyproject.toml` confiiguration and removing `--ignore-missing-imports` altogether as it can lead to issues.

Instead, I would suggest installing additional types if needed (like `pandas-stubs`, `types-psutil`, ...) to your `test` feature next to `mypy` and for dependencies that don't have stubs yet adding an exception to `pyproject.toml` as follows:

```toml
[[tool.mypy.overrides]]
# https://github.com/scikit-learn/scikit-learn/issues/16705
module = ["sklearn.*"]
ignore_missing_imports = true
```

stanmart:
Thanks, @MarcAntoineSchmidtQC and @pavelzw! I think most of your comments are now addressed. Later this week I'll add the type stub packages for dependencies that have them, and then try removing `ignore_missing_imports`.

MarcAntoineSchmidtQC:
> Thanks, @MarcAntoineSchmidtQC and @pavelzw! I think most of your comments are now addressed. Later this week I'll add the type stub packages for dependencies that have them, and then try removing `ignore_missing_imports`.

This seems to be rather tedious (I tried removing the flag and there are a lot of missing packages) because most of them will need there specific ignore config. I would prefer if we merged this first, and left this as a future task.

stanmart:
Sounds good! I'm that case this PR looks ready to merge to me. Maybe a final question: do we need a changelog entry for this?

jtilly:
Before merging this, we need to update the docs (e.g., https://glum.readthedocs.io/en/latest/contributing.html#install-for-development) and explain how to use this new machinery.

stanmart:
I added [some instructions](https://github.com/Quantco/glum/pull/802/files#diff-e052d37005ef738791e533f325afeb1f194d5d2c18e5cb0cd992ca56097baba0) a while age, but it can be expanded a bit (e.g. with how to [install both glum and tabmat in editable mode](https://github.com/Quantco/glum/pull/802#issuecomment-2150632124)).

docs/contributing.rst
```diff
@@ -31,39 +31,47 @@ Install for development
 --------------------------------------------------
 
 The first step is to set up a conda environment and install glum in editable mode.
+The project uses [pixi](https://pixi.sh/latest/) for managing the environment. If you don't have pixi installed, start by [installing it](https://pixi.sh/latest/#installation).
 We strongly suggest to use ``mamba`` instead of ``conda`` as this provides the same functionality at much greater speed.
```
jtilly, line 35:
```suggestion
```

jtilly (approved):

MarcAntoineSchmidtQC:
@stanmart, was there something else you wanted to do in this PR? Happy to merge whenever you are okay with it.

stanmart:
I think we can merge this as it is.

PR #841, opened 2024-09-18
Fix dtype mismatch in GLR(alpha_search=True).fit()

mlondschien:
```python
import numpy as np
import tabmat
import pandas as pd
import glum

n = 1_234
rng = np.random.default_rng(0)
p = 0.01

x = rng.choice([0, 1], p=[p, 1-p], size=n)
cat = pd.Categorical.from_codes(x, categories=np.array(range(3), dtype=np.int32))
mat = tabmat.CategoricalMatrix(cat, column_name="cat", dtype=np.float32)

y = np.random.choice([0, 1], size=n).astype(np.float32)

glm = glum.GeneralizedLinearRegressor(family="binomial", l1_ratio=0.5, alpha_search=True)
glm.fit(mat, y)
```
raises.


CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@
 Changelog
 =========
 
+3.0.3 - unreleased
+------------------
+
+**Bug fix:
+
+- Fixed a bug where :meth:`~glum.GeneralizedLinearRegressor.fit` would raise a `dtype` mismatch error if fit with `alpha_search=True`.
```
lbittarello, line 15:
```suggestion
- Fixed a bug where :meth:`glum.GeneralizedLinearRegressor.fit` would raise a ``dtype`` mismatch error if fit with ``alpha_search=True``.
```

lbittarello (approved):

PR #854, opened 2024-10-05
Drop SciPy from nightly CI

### LINKED ISSUE #853: Daily run failure: Unit tests
The daily unit tests failed. See https://github.com/Quantco/glum/actions/runs/11190590913 for details.

lbittarello:
Closes #853.

Is there a reason for which we aren't testing against nightly NumPy?


stanmart:
I don't know of any. It might be a good idea to add nightly numpy.

I know that these daily failures are super annoying, but the scipy nightly test can occasionally be useful (e.g. #811) Maybe we could raise the missing wheel issue upstream and re-add scipy if/when it is fixed.

stanmart (approved):

Issue #850, opened 2024-09-26
Allow setting `max_inner_iter` passed to `enet_coordinate_descent_gram`


mlondschien:
This is set [here](https://github.com/Quantco/glum/blob/2649d5b4abc8f9ae56a840975086f65689c8cad0/src/glum/_solvers.py#L424). Reducing this to a more reasonable number such as 1000 results in much faster convergence and fewer warnings

```
/cluster/customapps/biomed/grlab/users/lmalte/mambaforge/envs/icufm/lib/python3.10/site-packages/glum/_solvers.py:58: ConvergenceWarning: Coordinate descent did not converge. You might want to increase the number of iterations. Minimum norm subgradient: nan, tolerance: 0.0010000000474974513
  new_coef, gap, _, _, n_cycles = enet_coordinate_descent_gram(1036083102226]
```
although many more warnings
```
/Users/mlondschien/code/glum/src/glum/_solvers.py:58: ConvergenceWarning: Coordinate descent did not converge. You might want to increase the number of iterations. Minimum norm subgradient: 0.04440617933869362, tolerance: 0.0061285351403057575
  new_coef, gap, _, _, n_cycles = enet_coordinate_descent_gram(
```
(0.04 is much better than `nan`).

It would be nice to figure out why `enet_coordinate_descent_gram` does not converge in the insane amount of iterations `max_inner_iter=100_000` but rather returns infinite values. Shouldn't the hessian be positive definite, in particular, if `P1 != 0`?

PR #865, opened 2024-10-24
New argument: `max_inner_iter`

mlondschien:
contributes to https://github.com/Quantco/glum/issues/850


CHANGELOG.rst
```diff
@@ -7,9 +7,13 @@
 Changelog
 =========
 
-3.0.3 - unreleased
+3.1.0 - unreleased
 ------------------
 
+**New features:**
```
lbittarello, line 13:
```suggestion
**New feature:**
```

CHANGELOG.rst
```diff
@@ -7,9 +7,13 @@
...
+
+- New argument `max_inner_iter` for classes :class:`~glum.GeneralizedLinearRegressor` and :class:`~glum.GeneralizedLinearRegressorCV` to control the maximum number of iterations of the inner solver in the IRLS-CD algorithm.
```
lbittarello, line 15:
```suggestion
- New argument ``max_inner_iter`` for classes :class:`~glum.GeneralizedLinearRegressor` and :class:`~glum.GeneralizedLinearRegressorCV` to control the maximum number of iterations of the inner solver in the IRLS-CD algorithm.
```

lbittarello (approved):

MarcAntoineSchmidtQC (commented):
This is great, and was probably originally meant to be exposed. Thanks.

I'm wondering if we should change the default value to 1000 (as mentioned in the linked issue). Can you provide a simulation to show the difference in runtime between the two?

I'm happy to merge this without the change and switch the discussion to another issue.

mlondschien:
> Can you provide a simulation to show the difference in runtime between the two?

I have not managed to reproduce the convergence issues I encountered on a simulated dataset. The dataset I am using is an aggregation of ICU benchmarks. A training and approval by the dataset owner is required to access. So while not proprietary, it's not easy to share.

The defining feature of this dataset is that there are many strongly correlated features and many mean-imputed missing values.

lbittarello:
@stanmart, ping

stanmart (approved):
LGTM, but maybe don't close #850 until we test if a lower default works better.

mlondschien:
> LGTM, but maybe don't close https://github.com/Quantco/glum/issues/850 until we test if a lower default works better.

My understanding is the the `max_iter_inner` is only needed if there is some problem with the inner solver, e.g., the problem is not strictly convex. Exposing `max_iter_inner` to the user to avoid the solver trying to find a unique minimum that does not exist for many, many iterations is a hotfix. It would be good to solve the underlying problem: Why does the inner solver not converge in say 10 iterations? https://github.com/Quantco/tabmat/pull/408 is a first step. Still, `glum` does not behave as expected for problems such as the one in https://github.com/Quantco/glum/issues/872. Why? Something with the hessian.

PR #868, opened 2024-10-26
Compatibility with narwhals-based tabmat

### LINKED ISSUE #867: Daily run failure: Unit tests
The daily unit tests failed. See https://github.com/Quantco/glum/actions/runs/11529092707 for details.

stanmart:
Checklist

Aims to fix #867.

- One part of the issue was that narwhals is not (yet) an explicit dependency of glum (which is fine IMO) and nightlies are installed with `--no-deps`. This PR adds narwhals to the nightlies environment, but not the main one (where will be pulled in as a transitive dependency).
- The other issue is caused by the fact that `tabmat.from_pandas/df` now does not accept dataframes which have non-unique column names. This was an unintentended consequence of switching to narwhals, but I think it's probably for the best. Unfortunately many of our golden master benchmarks do have duplicate columns. I think the best way to go would be to fix the benchmark datasets, but let me know if you'd rather allow duplicate column names (in which case the fix should be on the tabmat side).


stanmart:
@MarcAntoineSchmidtQC let me know your thoughts about the duplicate column name issue.

MarcAntoineSchmidtQC:
adding narwhals to the "build-tabmat" group is good with me.

Regarding the duplicate columns, I think this was an issue with some of the logic. I agree that we should fix the benchmark dataset to not have duplicate names. We should also verify that we are giving an informative error message when this happens. 

Thanks for taking care of this!

stanmart:
> We should also verify that we are giving an informative error message when this happens.

We should do this in tabmat, right?

stanmart:
Thanks for the amazingly quick update @MarcoGorelli! 🙏  @MarcAntoineSchmidtQC, we can then merge this as is, right?

jtilly (approved):

PR #873, opened 2024-10-30
Run the CI unit tests on external pull requests

stanmart:
Checklist


stanmart:
Eh, this is not the solution because it will duplicate checks. What about only running the CI on pull requests and pushes to main? Although that would mean that working branches do not have CI. Or maybe leave everything as is and remember to check the source repo of external PRs? :)

jtilly:
Maybe imitate https://github.com/Quantco/metalearners/blob/a4858da69a9c8b83f2c460fa0240529c75e4f6c2/.github/workflows/ci.yml?

stanmart:
Thanks, that looks good. I'll even steal Kevin's detailed comments.

The slight issue remains that the CI won't run until we open a PR, but that's fine with me.

jtilly (approved):

PR #876, opened 2024-11-01
Add tutorial on how to estimate a Cox Proportional Hazards Model in glum

MatthiasSchmidtblaicherQC:
And: Add link to Matt Mills' tutorial on fitting penalized splines in glum.

Checklist


docs/tutorials/cox_model/cox_model.ipynb
```diff
@@ -0,0 +1,893 @@
...
+    "\n",
+    "## Table of Contents\n",
+    "* [1. Equivalence Between the Cox Likelihood and a Profile Poisson Likelihood](#1.-Equivalence-Between-the-Cox-Likelihood-and-a-Profile-Poisson-Likelihood)\n",
+    "* [2. Estimating a Cox Model in Glum](#2.-Estimating-a-Cox-Model-in-Glum)\n",
+    "* [3. Speed Considerations](#3.-Speed-Considerations)\n",
+    "\n",
+    "## 1. Equivalence Between the Cox Likelihood and a Profile Poisson Likelihood<a class=\"anchor\"></a>\n",
+    "\n",
+    "In the Cox model, the rate of event occurrence, $\\lambda(t,x_i)$, factorizes nicely into a linear predictor $\\eta_i=\\sum_k \\beta_k x_{ik}$ that depends on individual $i$'s characteristics but not on time $t$ and a baseline hazard $\\lambda_0$ that depends only on time, $\\lambda(t,x_i)=\\lambda_0(t)\\exp(\\eta_i)$ (the proportional hazards assumption). The partial log-likelihood of $\\eta_i$ is\n",
```
stanmart, line 20:
```suggestion
    "In the Cox model, the rate of event occurrence, $\\lambda(t,x_i)$, factorizes nicely into a linear predictor $\\eta_i=\\sum_k \\beta_k x_{ik}$ that depends on individual $i$'s characteristics but not on time $t$, and a baseline hazard $\\lambda_0$ that depends only on time: $\\lambda(t,x_i)=\\lambda_0(t)\\exp(\\eta_i)$. This is known as the proportional hazards assumption). The partial log-likelihood of $\\eta_i$ is\n",
```

docs/tutorials/cox_model/cox_model.ipynb
```diff
@@ -0,0 +1,893 @@
...
+    "$$\n",
+    "which is the same as the partial likelihood in the Cox model, apart from the -1 which drops out when taking derivatives. In short, the Cox partial log likelihood is equivalent to a Poisson log likelihood with the estimate for time period effects fed back in (\"profiled out\"). This means that, to estimate the parameters of the Cox model, one can simply run a Poisson regression with time fixed effects $\\alpha_t$.\n",
+    "\n",
+    "## 2. Estimating a Cox Model in Glum<a class=\"anchor\"></a>\n",
+    "\n",
+    "We now show that a Poisson approach in `glum` yields the same parameter estimates as a Cox model. For the latter, we use the [lifelines](https://github.com/CamDavidsonPilon/lifelines) library. We also take the dataset from lifelines, which is from an RCT on recidivism for 432 convicts released from Maryland state prisons with first arrest after release as event. We first load imports and the dataset. The dataset has one row per convict, with two outcome columns, the `week` until which the observation lasts and `arrest`, which indicates whether an arrest event happened or not (censoring)."
```
stanmart, line 40:
```suggestion
    "We now show that a Poisson approach in `glum` yields the same parameter estimates as a Cox model. For the latter, we use the [lifelines](https://github.com/CamDavidsonPilon/lifelines) library. We also take the dataset from lifelines, which is from an RCT on recidivism for 432 convicts released from Maryland state prisons with first arrest after release as event. We first load imports and the dataset. The dataset has one row per convict, with two outcome columns: the `week` until which the observation lasts and `arrest`, which indicates whether an arrest event happened or not (censoring)."
```

docs/tutorials/cox_model/cox_model.ipynb
```diff
@@ -0,0 +1,893 @@
...
+    "%timeit model_glum = glum.GeneralizedLinearRegressor(family=\"poisson\",formula=\"arrest ~ fin + bs(age, df=4) + race + wexp + mar + paro + bs(prio,df=3) + C(week)\").fit(survival_split(df, 'week', 'arrest'))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Footnotes\n",
+    "<span id=\"fn1\"> 1</span>:\n",
+    "The Cox model assumes that at most one individual has an event at any time (\"no ties\").\n",
```
stanmart, line 853: Maybe mention here that there are various ways to deal with this? (e.g., exact - and expensive - calculations, efron approximation (reasonably accurate), breslow approximation (fastest). Also, in the case of many ties, an inherently discrete survival model might be the best option.

docs/tutorials/cox_model/cox_model.ipynb
```diff
@@ -0,0 +1,893 @@
...
+    "$$\n",
+    "with cumulant function $b$ and dispersion parameter $\\sigma_i$. The linear predictor is given by $\\eta_i=g((b'(\\theta_i))$ for link function $g$. The Cox partial log-likelihood cannot be broken down into such a sum over individuals $i$.\n",
+    "\n",
+    "## References\n",
+    "\n",
+    "[1] Carstensen, B. 2023. Who needs the Cox model anyway. December. Available at: http://bendixcarstensen.com/WntCma.pdf.\n",
+    "\n",
+    "[2] Whitehead, J., 1980. Fitting Cox’s regression model to survival data using GLIM. _Journal of the Royal Statistical Society Series C: Applied Statistics_, 29(3), pp.268-275.\n",
+    "\n",
+    "[3] Zhong, C. and Tibshirani, R., 2019. Survival analysis as a classification problem. arXiv preprint arXiv:1909.11171."
```
stanmart, line 868: They do use the same dataset-transformation (which they call stacking) idea, but they rely on a binomial model instead of Poisson and say that it is a good enough approximation. So relevant, but I would not say it's the same thing.

stanmart (approved):
Thank you Matthias, I like this tutorial a lot!

Maybe we could also mention that the model we estimate is conceptually a piecewise constant hazard model, but because we have a different constant for each event time, it becomes essentially non-parametric. As a corollary, if one wants to reduce the number of parameters, they can opt for merging some categories and estimating fewer "constants".

docs/tutorials/cox_model/cox_model.ipynb
MatthiasSchmidtblaicherQC: Right. I removed that reference.

MatthiasSchmidtblaicherQC:
> Thank you Matthias, I like this tutorial a lot!
> 
> Maybe we could also mention that the model we estimate is conceptually a piecewise constant hazard model, but because we have a different constant for each event time, it becomes essentially non-parametric. As a corollary, if one wants to reduce the number of parameters, they can opt for merging some categories and estimating fewer "constants".

Thanks! I incorporated your suggestion and added a small conclusion section describing how one might want to depart from the Cox model. Feel free to have another look :).

stanmart (approved):

PR #881, opened 2024-11-08
MNT use dataclass for IRLSData

lorentzenchr:
A small neat maintenance PR using a `dataclass` decorator for `IRLSData`.


MarcAntoineSchmidtQC (approved):
Thanks for the cleanup.

Issue #888, opened 2024-12-12
Incorrect use of partial in function `TweedieDistribution._rowwise_gradient_hessian`


bili2002:
### Description
There is an error in the [code](https://github.com/Quantco/glum/blob/dc5421c13ebd590dad101967e6cd6e1c7053e7c0/src/glum/_distribution.py#L674-L675) for `TweedieDistribution._rowwise_gradient_hessian` when `self.power = 3`. A `p` parameter is passed to `partial`, however `p` is not defined in the [function signature](https://github.com/Quantco/glum/blob/dc5421c13ebd590dad101967e6cd6e1c7053e7c0/src/glum/_functions.pyx#L261-L268) for `inv_gaussian_log_rowwise_gradient_hessian`.

PR #889, opened 2024-12-23
Fix: Incorrect use of partial in `TweedieDistribution._rowwise_gradient_hessian`

bili2002:
This PR addresses issue #888 by fixing the incorrect use of partial in the `TweedieDistribution._rowwise_gradient_hessian` function.

Checklist


stanmart (changes requested):
Thank you for finding fixing this! I believe there is a similar mistake here: https://github.com/Quantco/glum/blob/9c0a22185f4ad7583302e107a6a4867ac6410334/src/glum/_distribution.py#L706
Could you please address it too so that they are in the same PR?

Also, I think it'd be worth adding a test case that would have spotted these bugs (maybe [here](https://github.com/Quantco/glum/blob/9c0a22185f4ad7583302e107a6a4867ac6410334/tests/glm/test_distribution.py#L167-L176)?), but I'm also o kay with doing it later.

stanmart:
Hm, it might be a bit more involved than that. Is it possible that `inv_gaussian_log_rowwise_gradient_hessian` is incorrect? Shouldn't it be

```diff
@@ -276,12 +276,12 @@ def inv_gaussian_log_rowwise_gradient_hessian(
         inv_mu = 1 / mu[i]
         inv_mu2 = inv_mu ** 2
 
-        gradient_rows_out[i] = 2 * weights[i] * (inv_mu - y[i] * inv_mu2)
-        hessian_rows_out[i] = 2 * weights[i] * (inv_mu - 2 * y[i] * inv_mu2)
+        gradient_rows_out[i] = weights[i] * (y[i] * inv_mu2 - inv_mu)
+        hessian_rows_out[i] = weights[i] * (2 * y[i] * inv_mu2 - inv_mu)
```
instead?

stanmart (approved):
Thank you :)

stanmart:
@bili2002, do you mind if I push the fix for those formulas to this branch before merging this PR?

bili2002:
@stanmart Sure, go ahead with the formulas!

stanmart:
@lbittarello, could you please check if my changes make sense?

stanmart:
Also, if it's correct, I  believe the [relevant part](https://github.com/Quantco/objectives/blob/06823e369680a635f10e6689b57d4db547015632/src/functions/inverse_gaussian.rs#L34-L41) of quantco/objectives needs to be updated, too.

CHANGELOG.rst
```diff
@@ -10,6 +10,11 @@ Changelog
 UNRELEASED
 ----------
 
+**Bug fix:
+
+- Fixed a bug where :meth:`~glum.TweedieDistribution._rowwise_gradient_hessian` and :meth:`~glum.TweedieDistribution._eta_mu_deviance` would call functions with wrong arguments in the p = 3 case.
+- Fixed :class:`glum.InverseGaussianDistribution` not using the optimized gradient, hessian and deviance implementations, as well as those gradients and hessians being incorrect.
```
lbittarello, line 16:
```suggestion
- Fixed a bug where :meth:`~glum.TweedieDistribution._rowwise_gradient_hessian` and :meth:`~glum.TweedieDistribution._eta_mu_deviance` would call functions with wrong arguments in the ``p = 3`` case.
- Fixed :class:`glum.InverseGaussianDistribution` not using the optimized gradient, Hessian and deviance implementations, as well as those derivatives having the wrong sign.
```

src/glum/_functions.pyx
```diff
@@ -276,8 +276,8 @@ def inv_gaussian_log_rowwise_gradient_hessian(
         inv_mu = 1 / mu[i]
         inv_mu2 = inv_mu ** 2
 
-        gradient_rows_out[i] = 2 * weights[i] * (inv_mu - y[i] * inv_mu2)
-        hessian_rows_out[i] = 2 * weights[i] * (inv_mu - 2 * y[i] * inv_mu2)
+        gradient_rows_out[i] = weights[i] * (y[i] * inv_mu2 - inv_mu)
+        hessian_rows_out[i] = weights[i] * inv_mu
```
lbittarello, line 280:
```suggestion
        hessian_rows_out[i] = weights[i] * (2 * y[i] * inv_mu2 - inv_mu)
```

stanmart: I thought this only works for tweedies with 1 ≤ p ≤ 2, because otherwise the hessian is not positive definite ([see reference](https://github.com/Quantco/glum/blob/27b123d08f5aee157b0bc1cf3357a82cba3cc512/src/glum/_functions.pyx#L359-L361))

stanmart: Added a comment. I believe [this test case](https://github.com/Quantco/glum/blob/6f18433990bebae63af8b706ded803861819215d/tests/glm/test_distribution.py#L230) should cover the Hessian, or did you mean some other kind of test?

lbittarello: I was rather thinking of a test of a fitted model against some benchmark (e.g., `glmnet`).

stanmart: Ah, right, I can add a test case for that.

stanmart: Something like [this](https://github.com/Quantco/glum/blob/b67feb429892d81c3048a09bc9257fcea2de3894/tests/glm/test_glm.py#L1273-L1344)?

tests/glm/test_glm.py
```diff
@@ -1270,6 +1270,80 @@ def test_binomial_cloglog_unregularized(solver):
...
+        family="inverse.gaussian",
+        link="log",
+        solver="irls-cd",
+        gradient_tol=1e-8,
+        selection="random",
+        random_state=rng,
+    )
+    glm.fit(X, y)
+    assert_allclose(glm.intercept_, glmnet_intercept, rtol=1e-3)
+    assert_allclose(glm.coef_, glmnet_coef, rtol=1e-3)
```
lbittarello, line 1309: Out of curiosity, what happens if you use the full Hessian? Do the tests pass? Are the coefficients closer to `glmnet`'s?

lbittarello (approved):

tests/glm/test_glm.py
stanmart: The glmnet test passes and the coefficients are roughly as close as in the other case. Another test (`tests/glm/test_glm.py::test_glm_family_argument[inverse.gaussian-fam4]`) fails with a singular matrix `LinalgError`, though.

stanmart: The failing test might not be an issue though as I believe the model was originally meant to be penalized, which but it is not since glum v3. 🤔 Anyways, I think it needs a bit more thinking and testing if we want to go with the true Hessian. I'd vote for merging this fix relatively soon, and possibly switching to the true Hessian later on if we convince ourselves that it works.

PR #904, opened 2025-01-30
Reorganize the code base

lbittarello:
This PR breaks some files into smaller ones. It doesn't introduce any substantive changes.

Functions that are imported by another file have been renamed so they don't start with a leading underscore.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -52,7 +52,7 @@ jobs:
     if: github.event_name == 'release' && github.event.action == 'published'
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
-    environment: 
+    environment:
       name: test_release
       url: https://test.pypi.org/p/glum
     permissions:
@@ -70,7 +70,7 @@ jobs:
     if: github.event_name == 'release' && github.event.action == 'published'
     needs: [build_wheels, build_sdist, upload_testpypi]
     runs-on: ubuntu-latest
-    environment: 
+    environment:
       name: release
       url: https://pypi.org/p/glum
     permissions:
--- a/src/glum/_algebra.py
+++ b/src/glum/_algebra.py
@@ -1,0 +1,93 @@
+import typing
+from typing import Union
+
+import numpy as np
+import tabmat as tm
+from scipy import linalg, sparse
+
+
+def is_pos_semidef(p: Union[sparse.spmatrix, np.ndarray]) -> Union[bool, np.bool_]:
+    """
+    Checks for positive semidefiniteness of ``p`` if ``p`` is a matrix, or
+    ``diag(p)`` if a vector.
+
+    ``np.linalg.cholesky(P2)`` 'only' asserts positive definiteness; due to
+    numerical precision, we allow eigenvalues to be a tiny bit negative.
+    """
+    # 1d case
+    if p.ndim == 1 or p.shape[0] == 1:
+        any_negative = (p < 0).max() if sparse.isspmatrix(p) else (p < 0).any()
+        return not any_negative
+
+    # 2d case
+    # About -6e-7 for 32-bit, -1e-15 for 64-bit
+    epsneg = -10 * np.finfo(np.result_type(float, p.dtype)).epsneg
+
+    if sparse.issparse(p):
+        # Computing eigenvalues for sparse matrices is inefficient. If the matrix is
+        # not huge, convert to dense. Otherwise, calculate 10% of its eigenvalues.
+        p = typing.cast(sparse.spmatrix, p)
+        if p.shape[0] < 2000:
+            eigenvalues = linalg.eigvalsh(p.toarray())
+        else:
+            n_evals_to_compuate = p.shape[0] // 10 + 1
+            sigma = -1000 * epsneg  # start searching near this value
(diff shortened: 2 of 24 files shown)


src/glum/_glm_regressor.py
stanmart, line 1: Why not simply `_glm_py`, and the current `_glm.py` could become `_glm_base.py`. That would be a bit more aligned with class names IMO.

src/glum/_algebra.py
stanmart, line 1: What about `_linalg.py`? That would resemble numpy and scipy's submodules.

.github/workflows/build_wheels.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     if: github.event_name == 'release' && github.event.action == 'published'
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
-    environment:
```
stanmart, line 55: Should we add the trailing whitespace fixer pre-commit-hook so that these don't reappear?

src/glum/_glm.py
```diff
@@ -1,47 +1,19 @@
...
-import scipy.sparse.linalg as splinalg
 import sklearn as skl
 import tabmat as tm
-from formulaic import Formula, FormulaSpec
-from formulaic.parser import DefaultFormulaParser
-from formulaic.utils.constraints import LinearConstraintParser
-from formulaic.utils.context import capture_context
-from packaging import version
 from scipy import linalg, sparse, stats
-from sklearn.base import BaseEstimator, RegressorMixin
```
stanmart, line 37: I think these imports are quite idiomatic, and that's how they appear in scikit-learn's docs, too.

tests/glm/test_glm_cv.py
```diff
@@ -37,7 +36,7 @@ def test_normal_elastic_net_comparison(l1_ratio, fit_intercept, convert_x_fn):
     tol = 1e-9
 
     n_predict = 10
-    X, y, _ = make_regression(
+    X, y, _ = skl.datasets.make_regression(
```
stanmart, line 39: On the other hand I like using the qualified classname here as `make_regression` sounds like it could come from anywhere.

tests/glm/test_utils.py
```diff
@@ -21,7 +21,7 @@ def df():
     )
 
 
-def test_align_df_categories_numeric(df):
+def testalign_df_categories_numeric(df):
```
stanmart, line 24:
```suggestion
def test_align_df_categories_numeric(df):
```

And elsewhere. I imagine it's a typo.

tests/glm/test_utils.py
```diff
@@ -180,7 +180,7 @@ def df_na():
     )
 
 
-def test_add_missing_categories(df_na):
+def testadd_missing_categories(df_na):
```
stanmart, line 183:
```suggestion
def test_add_missing_categories(df_na):
```

stanmart (commented):
Thank you, I think the new structure is much nicer than what we had before. `_glm.py` in particular was rather unwieldy. I just have a few small comments, plus the following bigger ones.

> Functions that are imported by another file have been renamed so they don't start with a leading underscore.

Where does that convention come from? I though we mostly use underscores to denote functions which are private _for the users of glum_. I understand that these functions live in underscored modules, which can mean the same thing, but why switch? I personally don't see a huge benefit in using names to denote whether functions are module-private or package-private. When working on glum, it's simple to use 'Go to references` and `Rename symbol` to make changes across files anyways.

This is not to say that I'm totally opposed to the proposed change. If the consensus is to follow the new convention, that's fine with me.

> This PR breaks some files into smaller ones. It doesn't introduce any substantive changes.

This is my more important point: we should still check downstream projects before merging this to make sure we don't break to much stuff the this refactor (or at least give them some time for them to prepare for it). I know that you are not changing the public API, but it would not surprise me if downstream users were relying on some private functions, too.

lbittarello:
> Where does that convention come from? I though we mostly use underscores to denote functions which are private _for the users of glum_. I understand that these functions live in underscored modules [...], but why switch? I personally don't see a huge benefit in using names to denote whether functions are module-private or package-private.

At the moment, there's no pattern to which functions take a leading underscore. Some private functions don't have one (e.g., `is_pos_semidef`), some do. I tried to be more systematic. The advantage of this convention is that you can immediately tell whether other modules call a function or not when reading a file. We could also just use trailing underscores everywhere though.

> Why not simply `_glm_py`, and the current `_glm.py` could become `_glm_base.py`. That would be a bit more aligned with class names IMO.

It would, but I'm trying to minimise downstream breakage.

stanmart:
> At the moment, there's no pattern to which functions take a leading underscore. [...]

In that case I'm fine with either, and agree that we should pick one and be consistent.

stanmart:
Could [this](https://docs.python.org/3/library/doctest.html#doctest.NORMALIZE_WHITESPACE) solve the doctest issue?

lbittarello:
> Could [this](https://docs.python.org/3/library/doctest.html#doctest.NORMALIZE_WHITESPACE) solve the doctest issue?

It didn't.

stanmart (approved):
LGTM but I imagine @jtilly and @MarcAntoineSchmidtQC have more to say

PR #908, opened 2025-02-06
Lower absolute tolerance in _find_alpha_index

mlondschien:
Sometimes I use really small values for `alpha`. The new `atol` is arbitrary. Is should be <= 1e-10 for `float64`.

```python
In [2]: np.finfo(np.float32).eps
Out[2]: 1.1920929e-07

In [3]: np.finfo(np.float64).eps
Out[3]: 2.220446049250313e-16
```

Checklist

image

### DIFF
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -1279,7 +1279,8 @@ class GeneralizedLinearRegressorBase(BaseEstimator, RegressorMixin):
         if not self.alpha_search:
             raise ValueError
         # `np.isclose` because comparing floats is difficult
-        isclose = np.isclose(self._alphas, alpha)
+        atol = 100.0 * np.finfo(alpha.dtype).eps
+        isclose = np.isclose(self._alphas, alpha, atol=atol)
         if np.sum(isclose) == 1:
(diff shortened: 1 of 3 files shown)


lbittarello:
@MarcAntoineSchmidtQC, ping

src/glum/_glm.py
```diff
@@ -1279,7 +1279,8 @@ def _find_alpha_index(self, alpha):
         if not self.alpha_search:
             raise ValueError
         # `np.isclose` because comparing floats is difficult
-        isclose = np.isclose(self._alphas, alpha)
+        atol = 100.0 * np.finfo(alpha.dtype).eps
```
MarcAntoineSchmidtQC, line 1282: `alpha` is a scalar (e.g. float or np.float32), so it doesn't have a `dtype` attribute.
```suggestion
        atol = 100.0 * np.finfo(type(alpha)).eps
```

mlondschien:
#923 looks good to me. Maybe add a test with `alpha = 0`. This might fail if using only relative tolerance.

MarcAntoineSchmidtQC:
Closed in favor of #923

PR #911, opened 2025-02-17
Switch to rattler-build

pavelzw:
xref https://github.com/conda-forge/glum-feedstock/pull/43


MarcAntoineSchmidtQC:
I looked at the failure. Seems like there's an issue with rattler-build where it ignores the "target_platform" variant parameter, or it automatically modifies it. In the variant file of the failing build, we have `target_platform -> osx-64`, while rattler-build's solver indicates it's looking for `osx-arm64`. @pavelzw, should we open an issue on rattler-build?

pavelzw:
There already is one: https://github.com/prefix-dev/rattler-build/issues/282
We should pass --target-platform explicitly for now

MarcAntoineSchmidtQC (approved):

PR #944, opened 2025-09-08
Set `object_as_cat` to `True` when constructing the `tabmat` matrix

### LINKED ISSUE #939: Daily run failure: Unit tests
The daily unit tests failed. See https://github.com/Quantco/glum/actions/runs/15951749070 for details.

jtilly:
Fixes #939.

`formulaic` isn't compatible with the `str` dtype that becomes the default in `pandas>=3`. There's an open PR https://github.com/matthewwardrop/formulaic/pull/258, but not much work. My quick fix for this is to use categorical dtypes whenever we're encountering `objects` or `str` dtypes. Happy to entertain a different fix for this.

The relevant code snippet in tabmat is based on narwhals:
https://github.com/Quantco/tabmat/blob/fe2f30dd4c0b697b7bdfa9bf3c97065393d7d36a/src/tabmat/constructor.py#L96-L104

### DIFF
--- a/src/glum/_glm.py
+++ b/src/glum/_glm.py
@@ -281,6 +281,9 @@ class GeneralizedLinearRegressorBase(skl.base.RegressorMixin, skl.base.BaseEstim
                 self, "categorical_format", "{name}__{category}"
             ),
             cat_missing_method=cat_missing_method_after_alignment,
+            # setting object_as_cat to True allows us to bypass issues with the
+            # new str dtype in pandas 3
+            object_as_cat=True,
         )
 
         return X
@@ -1827,6 +1830,9 @@ class GeneralizedLinearRegressorBase(skl.base.RegressorMixin, skl.base.BaseEstim
                     ),
                     cat_missing_method=getattr(self, "cat_missing_method", "fail"),
                     cat_missing_name=getattr(self, "cat_missing_name", "(MISSING)"),
+                    # setting object_as_cat to True allows us to bypass issues with the
+                    # new str dtype in pandas 3
+                    object_as_cat=True,
                 )
 
         if y is None:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,15 @@
 Changelog
 =========
 
+3.1.3 - 2025-09-09
+------------------
+
+**Bug fix**
+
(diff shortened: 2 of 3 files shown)


src/glum/_glm.py
```diff
@@ -281,6 +281,9 @@ def _convert_from_pandas(
                 self, "categorical_format", "{name}__{category}"
             ),
             cat_missing_method=cat_missing_method_after_alignment,
+            # setting object_as_cat to True allows us to bypass issues with the
```
stanmart, line 284: I believe this should be `from_df` as we deprecated `from_pandas`.

stanmart (approved):
Thank you! I'm happy with handling it here, we can implement and we can implement a solution in `tabmat` when the `formulaic` PR is merged. My only question is that can this change break some workflows? Object columns, which were dropped before, will now be treated as categoricals. Or does relying on that fall into [this category](https://xkcd.com/1172/)?

stanmart:
Yeah I was a bit confused because I also thought they were not, but here's what I get:

```python
import numpy as np
import pandas as pd

df = pd.DataFrame(
    {
        "x": np.random.rand(10),
        "s": np.random.choice(["a", "b"], size=10),
        "y": np.random.rand(10),
    }
)

model = GeneralizedLinearRegressor()
print(model.fit(df[["x", "s"]], df["y"]).coef_table())
# /Users/stanmart/work/glum/.pixi/envs/default/lib/python3.13/site-packages/tabmat/constructor.py:179: UserWarning: Columns ['s'] were ignored. Make sure they have a valid dtype.
#   warnings.warn(
# intercept    0.253587
# x            0.284256
# Name: coef, dtype: float64 
```

They are not dropped when using a formula, though:

```python
model_f = GeneralizedLinearRegressor(formula="y ~ x + s", drop_first=True)
print(model_f.fit(df).coef_table())
# intercept    0.315461
# x            0.578222
# s[b]        -0.028602
# Name: coef, dtype: float64
```

So I'm okay with making it more consistent and always treating objects as categoricals.

jtilly:
> By fix, do you mean making formulas also drop object columns?

No. Just convert `str` dtypes when using formulaic.

jtilly:
Let's address this in tabmat instead.

PR #996, opened 2026-04-19
Monotonic constraints

### LINKED ISSUE #882: Splines with difference penalty and monotonic constraints
As a user of GLMs, I would like to
- splines with a difference penalty without the hassle to specify the `P2` matrix myself
- optimize a single penalty per term (think of 2 features both modelled as splines)
- use splines with monotonicity constraints

About monotonic splines, cf. Ramsay "Monotone Spline Regressions in Action" https://doi.org/10.1214/ss/1177012761:
- One can use M- or I-spline bases; or
- Use differences of B-Splines
- One needs to restrict the coefficients to be either `>=0` (increasing) or `<=0` (decreasing)

MatthiasSchmidtblaicherQC:
### Motivation

Currently, the only way to enforce inequality constraints in glum is via `A_ineq`/`b_ineq` with `solver='trust-constr'` This required users to manually construct constraint matrices, and understand the column layout of the expanded design matrix. See also [#882.](https://github.com/Quantco/glum/issues/882)

In addition, the trust-constrained solver is slow on the insurance benchmark dataset and other realistic datasets (200-500 iterations, often failing to converge on gamma-family problems).

### What this PR adds

**Formula-level `monotonic_constraints` parameter.** Users pass a dictionary mapping variable names to `"increasing"` or `"decreasing"`:

```python
model = GeneralizedLinearRegressor(
    formula="y ~ bs(BonusMalus, df=10) + C(Region) + ...",
    monotonic_constraints={"BonusMalus": "increasing"},
)
model.fit(df)
```

This is an ergonomic improvement over manually building `A_ineq`/`b_ineq`:  users specify constraints on variable names not on individual columns. glum automatically resolves variable names to expanded formula terms (spline basis columns, ordered categoricals). Monotonic constraints also work for interacted variables.

The raw `A_ineq`/`b_ineq` interface remains available for users who want full control.

**New `irls-ls-monotonic` solver.** Inspired by [pygam's penalty-based approach](https://github.com/dswah/pyGAM), this is a penalty-based IRLS algorithm that adds a progressive quadratic penalty `λ · A_violated.T @ A_violated` to the Hessian for violated constraints. The penalty ramps from `λ=1e5` by ×100 per violation, capped at `1e9`. This reuses glum's existing fast IRLS infrastructure (`IRLSData`, `IRLSState`, line search, convergence checks) with little added complexity.

### Benchmark results 

Insurance benchmark problems (narrow/wide × gamma/poisson) with new monotonicity constraint on BonusMalus:

| Dataset | irls-ls-monotonic | trust-constr | Speedup |
|---|---|---|---|
| narrow-insurance gamma (n=25k, p=182) | **0.16s, 8 iter, converged** | 2.92s, 500 iter, not converged | 18× |
| narrow-insurance poisson (n=678k, p=225) | **0.88s, 11 iter, converged** | 4.64s, 207 iter, converged | 5× |
| wide-insurance gamma (n=25k, p=182) | **0.15s, 8 iter, converged** | 2.95s, 500 iter, not converged | 20× |
| wide-insurance poisson (n=678k, p=225) | **0.83s, 9 iter, converged** | 4.66s, 207 iter, converged | 6× |

Key takeaway: irls-ls-monotonic converges in 8-11 iterations on all problems. trust-constr fails to converge on gamma problems within 500 iterations.

Against pygam (the main Python alternative for shape-constrained GAMs), glum is 14-130× faster on the same problems, see the new benchmark figure.

**Synthetic benchmarks** showed that trust-constr is faster for small feature counts (p<50), but the crossover favors irls-ls-monotonic at the insurance-scale dimensions (p=150-225) that motivated this feature. At p=274, n=500k the two are comparable (~35s each). Users who prefer trust-constr for their problem shape can still set `solver='trust-constr'` explicitly.

**Overhead vs unconstrained IRLS:** monotonic constraints add ~2-4× overhead (6-11 iterations vs 2-3 iterations), and proportional slowdown compared to the unconstrained problem.

Checklist


src/glum/_formula.py
```diff
@@ -53,3 +58,161 @@ def parse_formula(
...
+_FACTOR_RE = re.compile(
+    r"""
+    (?P<name>[^:\[\]]+)    # factor name — everything up to ':', '[', or ']'
+    (?:\[(?P<idx>[^\]]*)\])?  # optional bracketed index
+    """,
+    re.VERBOSE,
+)
+
+
+def _parse_column_name(col: str, separator: str = ":") -> list[tuple[str, str | None]]:
```
MarcAntoineSchmidtQC, line 72: Is this the right place for the logic (all of `_formula.py`)? I might be wrong but everything here seems to be doing the job that formulaic should be doing. Formulaic probably doesn't have the capability to parse the exact constraints we want, but I want to understand whether there's a simpler way of handling this. It seems that we would have to refactor this whole logic if formulaic changed its API.

src/glum/_glm.py
```diff
@@ -1882,6 +1934,27 @@ def _should_copy_X(self):
         # If self.copy_X is False, check for data of wrong dtype and error if it exists.
         return self.copy_X or False
 
+    def _resolve_monotonic_constraints(
+        self,
+    ) -> tuple[Optional[np.ndarray], Optional[np.ndarray]]:
+        """Resolve ``monotonic_constraints`` to ``(A_ineq, b_ineq)``."""
+        if self.monotonic_constraints is None:
```
MarcAntoineSchmidtQC, line 1941: Can we check for None in the upstream context? I would prefer having something like:
```python
has_mono_c = self.monotonic_constraints is not None
has_a_b_ineq =  self.A_ineq is not None and self.B_ineq is not None
if has_mono_c and has_a_b_ineq:
    raise ...
elif has_mono_c and not has_a_b_ineq:
    _resolve_monotonic_constraints(...)
elif has_a_b_ineq and not has_mono_c:
    copy A_ineq and B_ineq
else:
    no constraints...
```

src/glum/_solvers.py
```diff
@@ -984,3 +984,146 @@ def _trust_constr_solver(
...
+        )
+
+    idx = 1 if data.fit_intercept else 0
+    constraint_lam = constraint_lam_start
+
+    P2_base = data.P2
+    if sparse.issparse(P2_base):
+        P2_base = P2_base.copy()
+    elif isinstance(P2_base, np.ndarray):
+        P2_base = P2_base.copy()
```
MarcAntoineSchmidtQC, line 1027: Seems like this conditional branching is unnecessary if we are doing the same thing in both cases.

src/glum/_solvers.py
```diff
@@ -984,3 +984,146 @@ def _trust_constr_solver(
...
+        C = _build_monotonic_penalty(coef[idx:], A_ineq, b_ineq, lam)
+        if C is None:
+            data.P2 = P2_base
+        else:
+            if P2_base.ndim == 1:
+                data.P2 = np.diag(P2_base) + C
+            elif sparse.issparse(P2_base):
+                data.P2 = P2_base.toarray() + C  # type: ignore[union-attr]
+            else:
+                data.P2 = P2_base + C
```
MarcAntoineSchmidtQC, line 1039: I think this will cause a lot of slowdown if we have many (implicit) columns (e.g. high-dimentional categorical). Is there a way to keep the sparcity of the data?
- `np.diag` will transform a 1-d array of size n into a nxn matrix.
- `.toarray()` is explicitly transforming the sparce array into dense.
Looking at `_build_monotonic_penalty`, it seems like all these operations could be done with a sparse array.

src/glum/_glm.py
```diff
@@ -597,11 +605,18 @@ def _solve(
...
+                    coef,
+                    self.n_iter_,
+                    self._n_cycles,
+                    self.diagnostics_,
+                ) = _monotonic_irls_solver(coef, irls_data, A_ineq, b_ineq)
+            elif self._solver == "irls-ls":
                 coef, self.n_iter_, self._n_cycles, self.diagnostics_ = _irls_solver(
                     _least_squares_solver, coef, irls_data
                 )
-            # 4.2 coordinate descent ##############################################
```
MarcAntoineSchmidtQC, line 604: Is there a reason to remove those comments (other than a cleanup)?  Same thing with lines 575 to 577.

src/glum/_glm.py
```diff
@@ -2690,8 +2767,10 @@ def fit(
         lower_bounds = check_bounds(self.lower_bounds, X.shape[1], X.dtype)
         upper_bounds = check_bounds(self.upper_bounds, X.shape[1], X.dtype)
 
+        _A = _mc_A_ineq if _mc_A_ineq is not None else self.A_ineq
```
MarcAntoineSchmidtQC, line 2770: Maybe we should raise ValueError if the user specifies both monotonic_constraints and A_ineq (or B_ineq). Currently, we are simply ignoring A_ineq.

MarcAntoineSchmidtQC (commented):
This is a great PR. My main concern is with the "densification" of the P2 matrix. I wonder whether it's necessary and if we could not make this faster for problem with high-dimentional categoricals.

src/glum/_formula.py
MatthiasSchmidtblaicherQC: Good question, but formulaic indeed does not supply all we need. We need a mapping from column index to position in each factor, which we don't get from there. I also tried an alternative approach that used more of formulaic's internals `term_indices`, etc., but that ended being more complicated and arguably more fragile.

Since are private functions, I would just leave them as is and change or remove them if formulaic changes if that is fine with you.

src/glum/_glm.py
MatthiasSchmidtblaicherQC: This was the AI assistant being overly eager when told to be terse, I added them back now :)

src/glum/_solvers.py
MatthiasSchmidtblaicherQC: Good point. I changed it accordingly such that _build_monotonic_penalty now returns a sparse.csc_matrix instead of a dense array.

We ran some benchmarks and the memory footprint improvement is big.

## Results
### Speed (50 iterations, monotonic constraints on spline + high-dim categorical)

| p | 1D | sparse | dense |
|---|---|---|---|
| ~500 | 0.55s | 0.55s | 0.57s |
| ~900 | 1.34s | 1.38s | 1.44s |
| ~1100 | 1.89s | 1.96s | 2.05s |

### Memory (P2 + constraint penalty, 5 violated constraints)

| p | sparse (`diags+C`) | dense (`np.diag+C`) |
|---|---|---|
| 500 | 7.9 KB | 1.9 MB |
| 2,000 | 31 KB | 30.5 MB |
| 10,000 | 156 KB | 763 MB |

<details>
    <summary>## Benchmarking code:</summary>
    ```python
"""Benchmark sparse vs dense P2 penalty with high-dimensional categoricals.

The monotonic solver builds P2 = P2_base + C where C is the constraint penalty.
Before the sparse fix, P2_base (1D diagonal or sparse) was expanded to dense via
np.diag() or .toarray(), creating an O(p^2) dense matrix even when p is large.
This benchmark measures the impact on a problem with many categorical levels.
"""
import time
import warnings
import numpy as np
from scipy import sparse as sp
from sklearn.preprocessing import OneHotEncoder, SplineTransformer

warnings.filterwarnings("ignore")
rng = np.random.default_rng(42)

configs = [
    {"n": 2000, "n_cats": 500, "label": "n=2k, p~510"},
    {"n": 2000, "n_cats": 1000, "label": "n=2k, p~1010"},
    {"n": 2000, "n_cats": 1500, "label": "n=2k, p~1510"},
]

for cfg in configs:
    n = cfg["n"]
    n_cats = cfg["n_cats"]

    # Continuous feature with spline basis (monotonic target)
    x = np.sort(rng.uniform(0, 1, n))
    sp_X = SplineTransformer(n_knots=5, degree=3, include_bias=False).fit_transform(
        x.reshape(-1, 1)
    )
    n_sp = sp_X.shape[1]

    # High-dimensional categorical
    cat = rng.integers(0, n_cats, n)
    ohe = OneHotEncoder(sparse_output=False, drop=None).fit_transform(cat.reshape(-1, 1))

    X = np.hstack([sp_X, ohe])
    p = X.shape[1]
    # Non-monotonic signal to force constraint violations
    y = (np.sin(4 * np.pi * x) + rng.standard_normal(n) * 0.3).clip(0.01)

    # Monotonic constraints only on the spline columns
    A_ineq = np.zeros((n_sp - 1, p))
    for i in range(n_sp - 1):
        A_ineq[i, i] = 1.0
        A_ineq[i, i + 1] = -1.0
    b_ineq = np.zeros(n_sp - 1)

    print(f"\n{cfg['label']} (actual p={p}, constraints={A_ineq.shape[0]})")
    print("-" * 60)

    for P2_type, P2_val in [
        ("1D (diagonal)", np.full(p, 0.01)),
        ("sparse", sp.eye(p, format="csc") * 0.01),
        ("dense", np.eye(p) * 0.01),
    ]:
        common = dict(
            A_ineq=A_ineq, b_ineq=b_ineq, alpha=1.0, l1_ratio=0,
            fit_intercept=False, gradient_tol=1e-8,
            solver="irls-ls-monotonic", max_iter=50,
        )
        times = []
        for _ in range(3):
            m = GeneralizedLinearRegressor(P2=P2_val, **common)
            t0 = time.perf_counter()
            m.fit(X, y)
            times.append(time.perf_counter() - t0)
        med = np.median(times)
        print(f"  P2={P2_type:15s}  {med:.3f}s  iter={m.n_iter_}")
```

</details>

MarcAntoineSchmidtQC (approved):
I like the new memory benchmarks. Ready to merge in my opinion. Thanks!

PR #999, opened 2026-04-28
Add test suite for backwards compatibility

jtilly:
I'm adding a test suite to ensure that we remain backwards compatible in the sense that we can still predict from models that were fitted using an old version of glum.

Fits a GLM with each released minor version of glum (discovered dynamically via `pixi search`), pickles the model, then unpickles it under the current version and checks that predictions match.                                                                                                                                                                        
                                                                                                                                                                                                                                                              
Three new files under `tests/backwards_compatibility/`:                                                                                                                                                                                                     
 - `fit.py` — fits and saves a model + predictions; accepts a version label or `HEAD`
 - `run.py` — orchestrates `fit.py` across all versions and compares predictions against HEAD                                                                                                                                                                
                                                                                                                                                                                                                                                              
Runs on PRs via a GitHub Actions workflow. Also available as `pixi run test-backwards-compatibility`.

### DIFF
--- a/.github/workflows/backwards-compatibility.yml
+++ b/.github/workflows/backwards-compatibility.yml
@@ -1,0 +1,35 @@
+name: Backwards Compatibility
+
+on:
+  # We would like to trigger for CI for any pull request action -
+  # both from QuantCo's branches as well as forks.
+  pull_request:
+  # In addition to pull requests, we want to run CI for pushes
+  # to the main branch and tags.
+  push:
+    branches:
+      - "main"
+    tags:
+      - "*"
+  workflow_dispatch:
+
+jobs:
+  backwards-compatibility:
+    name: Backwards compatibility tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set up pixi
+        uses: prefix-dev/setup-pixi@a0af7a228712d6121d37aba47adf55c1332c9c2e # v0.9.4
+        with:
+          environments: default
+          cache: true
+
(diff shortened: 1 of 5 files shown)


MarcAntoineSchmidtQC:
I like what this PR is doing, but I wonder if we need to create a dozen pixi environments every time. My understanding is that we want to ensure we are able to predict from an old model. Why not fit once and store the pickle file in git, then check in the current environment if we are able to load it.

MarcAntoineSchmidtQC:
Also, FYI for others that wants to test this locally. Older versions of pixi won't work. You need to update pixi (I had 0.63 and it didn't work).

stanmart:
> I like what this PR is doing, but I wonder if we need to create a dozen pixi environments every time. My understanding is that we want to ensure we are able to predict from an old model. Why not fit once and store the pickle file in git, then check in the current environment if we are able to load it.

We have something like that here, we can perhaps expand the related test suite: https://github.com/Quantco/glum/tree/a25bdcfc80a597102b0924929b956021dd6dd7f7/tests/glm/pickles

MarcAntoineSchmidtQC:
I just saw the comment Jan made on the quantcore.learn repo:

> [Marc-Antoine Schmidt (@MarcAntoineSchmidtQC)](https://github.com/MarcAntoineSchmidtQC) I would like to keep this consistent with glum, so let's agree on the setup here. As I write in the PR description, I'm also happy to check everything into git. The downside of that is that pickling/unpickling might fail due to other reasons, e.g. changes in pandas or scikit-learn versions. Recreating the environments is a bit more robust in that regard.

Making it more robust to changes in the dependencies is a good point. I will merge this.

@stanmart, I didn't remember this was added. We don't need it now but maybe for a future test it might be good to expand it.

MarcAntoineSchmidtQC (approved):

PR #1005, opened 2026-05-15
Enhance StepwiseGLM with caching, optimizations, and user guide

TylerGrantSmith:
Checklist


MarcAntoineSchmidtQC:
Thanks for contributing! I tested this PR and I'm not able to achieve significant speedups compared to an already existing option to create the TabMat matrix first, and using warm-start (maybe also using `tabmat.hstack`).

Maybe my benchmark was not using the type of problems this is intended to solve, so could you add more context and details, and maybe a benchmark with runtime.

TylerGrantSmith:
> Thanks for contributing! I tested this PR and I'm not able to achieve significant speedups compared to an already existing option to create the TabMat matrix first, and using warm-start (maybe also using `tabmat.hstack`).
> 
> Maybe my benchmark was not using the type of problems this is intended to solve, so could you add more context and details, and maybe a benchmark with runtime.

My apologies!  I had this set up on a forked repo and was not expecting it to show up here (yet).

